PR #121, opened 2018-11-05
properly build indices for single-coord dimensions

mschrimpf:
workaround to https://github.com/pydata/xarray/issues/2537


mschrimpf:
out-dated, maybe revisit later

PR #178, opened 2019-12-05
Not a number error fix

franzigeiger:
Pearson correlation returns a NaN value in case of only one value to correlate. For a small mock assembly I have only one test example and therefore the correlation threw an error. This fixes the problem.


mschrimpf:
That is the correct behavior for a correlation -- it's mathematically incorrect to correlate a single value! I think it would be better to have a test assembly with 2 values

PR #194, opened 2020-01-26
add visual search

shashikg:
check example [here](https://github.com/shashikg/brain-score/blob/master/examples/visual-search-benchmark.ipynb)


brainscore/benchmarks/__init__.py
```diff
@@ -91,6 +91,9 @@ def __init__(self):
         from .imagenet import Imagenet2012
         self['fei-fei.Deng2009-top1'] = LazyLoad(Imagenet2012)
 
+        from .search import KlabZhang2018ObjArray
+        self['klab.Zhang2018-ObjArray'] = LazyLoad(KlabZhang2018ObjArray)
```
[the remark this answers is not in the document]
shashikg: sure! I will do that

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
+import numpy as np
+import ScanMatchPy
```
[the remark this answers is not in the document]
shashikg: I forgot to update setup.py, will do that.
However, I have another doubt related to the ScanMatchPy package as this is the one which I created using the Matlab code. I will mail the detailed problem as I guess Mengmi also needs to look into it.

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        self._identifier='klab.Zhang2018-ObjArray'
+        self.parent='visual_search',
+        self.paper_link='https://doi.org/10.1038/s41467-018-06217-x'
+        self._assemblies = brainscore.get_assembly('klab.Zhang2018search_obj_array')
+        self._stimuli = self._assemblies.stimulus_set
+
+    def __call__(self, candidate: BrainModel):
+        self._metric = ScanMatchPy.initialize()
+
+        print("## Starting visual search task...")
```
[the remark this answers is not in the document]
shashikg: ok

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        candidate.start_task(BrainModel.Task.visual_search)
+        self.cumm_perf, self.saccades = candidate.look_at(self._stimuli)
+        fix_model = self.saccades[:,:7,:]
+        I_fix_model = self.saccades[:,7,:1]
+        fix1 = matlab.int32(fix_model.tolist())
+        I_fix1 = matlab.int32(I_fix_model.tolist())
+        print("## Search task done...\n")
```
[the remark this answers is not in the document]
shashikg: ok

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        self._stimuli = self._assemblies.stimulus_set
+
+    def __call__(self, candidate: BrainModel):
+        self._metric = ScanMatchPy.initialize()
+
+        print("## Starting visual search task...")
+        candidate.start_task(BrainModel.Task.visual_search)
+        self.cumm_perf, self.saccades = candidate.look_at(self._stimuli)
+        fix_model = self.saccades[:,:7,:]
+        I_fix_model = self.saccades[:,7,:1]
```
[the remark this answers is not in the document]
shashikg: Sure, I will add proper comments to each part of the next time!
self .saccades' 2nd dims have a size of 8,  till 0:7 it contains the position of the saccade, whereas the 7th index contains the index value of when it found the target.

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        fix1 = matlab.int32(fix_model.tolist())
+        I_fix1 = matlab.int32(I_fix_model.tolist())
+        print("## Search task done...\n")
+
+        print("## Calculating score...")
```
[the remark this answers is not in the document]
shashikg: ok

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        scores = []
+        for sub_id in tqdm(range(15), desc="comparing with human data: "):
+            data_human = self._assemblies.values[sub_id*300:(sub_id+1)*300]
```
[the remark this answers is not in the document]
shashikg: I will comment on it.
btw, there are 15 subjects and each has 300 search data for those 300 stimuli in the object array search dataset.

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+        self.raw_score = np.mean(scores)
+        self.std = np.std(scores)/np.sqrt(scores.shape[0])
+
+        self.model_score = Score([self.raw_score, self.std],
+                        coords={'aggregation': ['center', 'error']}, dims=['aggregation'])
+
+        self._metric.terminate()
+
+        ceiled_score = self.ceiling
+        print("## Score calculated...\n")
```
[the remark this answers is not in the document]
shashikg: ok

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,74 @@
...
+    @property
+    def identifier(self):
+        return self._identifier
+
+    @property
+    def version(self):
+        return self._version
+
+    @property
+    def ceiling(self):
```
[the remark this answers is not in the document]
shashikg: I will look into it

tests/test_benchmarks/test_search.py
```diff
@@ -0,0 +1,20 @@
+from candidate_models.model_commitments import brain_translated_pool
```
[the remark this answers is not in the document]
shashikg: I will look into it

mschrimpf:
There's an error by `ScanMatchPy` (also see travis log): `RuntimeError: On Linux, you must set the environment variable "LD_LIBRARY_PATH" to a non-empty string. For more details, see the package documentation.`

shashikg:
Related to ScanMatchPy:
You will need MATLAB Runtime to be installed:
https://www.mathworks.com/help/compiler/install-the-matlab-runtime.html 

Usually, after installing the Matlab Runtime, you need to add Matlab runtime path the LD_LIBRARY_PATH variable which normally empty by default (probably this is the reason why it's showing the above error)

mschrimpf:
Hm is there an automated way of installing matlab (i.e. using only command line arguments)? Our unit tests always setup a blank VM, so everything needs to be installed in there (see e.g. this log https://travis-ci.com/brain-score/brain-score/jobs/291366786)

shashikg:
I'm not sure but maybe something can be done using shell scripting. Let me check that.

mschrimpf:
The best case scenario would be if we could add it to the `setup.py` dependencies. That file can execute any code, so in principle should be able to install matlab. There's this new Matlab<>Python engine, not sure if that one would be helpful

shashikg:
@mschrimpf do you know how to run bash script from `setup.py` ??

shashikg:
> @mschrimpf do you know how to run bash script from `setup.py` ??

@mschrimpf

mschrimpf:
@shashikg sorry for the delay, things have been a little crazy with COVID19.
Since `setup.py` is just treated as a standard python script, you should be able to just use `subprocess`, e.g. `subprocess.call("script.sh")`

shashikg:
can u add the following lines to the Travis and Jenkins CI's shell before running the test examples?

`LD_LIBRARY_PATH=/usr/local/MATLAB/MATLAB_Runtime/v97/runtime/glnxa64:/usr/local/MATLAB/MATLAB_Runtime/v97/bin/glnxa64:/usr/local/MATLAB/MATLAB_Runtime/v97/sys/os/glnxa64:/usr/local/MATLAB/MATLAB_Runtime/v97/extern/bin/glnxa64:LD_LIBRARY_PATH`

`export LD_LIBRARY_PATH`

I added the script which will install the Matlab runtime while installing `ScanMatchPy` but you will need to modify the environment variables. I also added this thing in the shell script to add this to the `.bashrc` file but that will be only available for the end-users as when calling the script from `setup.py` script runs in a different shell. So this env variable will not be available for the shell which is running the build process.

mschrimpf:
Ok will do.

btw: I think you have an error in https://github.com/kreimanlab/ScanMatchPy/blob/58200136931be53d8ea9b38d99218b6f2973f625/install_matlab_runtime.sh#L15-L18 (lines 15 and 18): you're setting `LD_LIBRARY_PATH=...:LD_LIBRARY_PATH` but you're missing a `$`, i.e. `LD_LIBRARY_PATH=...:$LD_LIBRARY_PATH`. Otherwise it's going to set the literal string `LD_LIBRARY_PATH` instead of the referenced value.

shashikg:
Thanks for pointing out, will change that.

mschrimpf:
okay changed (at least for travis, jenkins we might soon be able to just use the `.travis.yml` as well). I also merged `master` into your fork again.

btw, could we simplify the task identifier from `visual_search_obj_arr` to `visual_search`? (would have to be done here + in model-tools)

shashikg:
> okay changed (at least for travis, jenkins we might soon be able to just use the `.travis.yml` as well).
> image
> Seems like it installs okay now.
> I also merged `master` into your fork again.
> 
> btw, could we simplify the task identifier from `visual_search_obj_arr` to `visual_search`? (would have to be done here + in model-tools)

so actually there are three different tasks to do in visual search i.e. object_array, natural_images, and waldo. I guess @Mengmi can suggest what we should do here.

Mengmi:
> > btw, could we simplify the task identifier from `visual_search_obj_arr` to `visual_search`? (would have to be done here + in model-tools)
> 
> so actually there are three different tasks to do in visual search i.e. object_array, natural_images, and waldo. I guess @Mengmi can suggest what we should do here.

Thanks for making this happen, Shashi and Martin! To avoid confusion, I suggest that we use one identifier as visual search, then it automatically calls three evaluation functions on all three datasets respectively.

mschrimpf:
Do all three of these tasks use different mechanisms? I thought they would be based on the same "visual search" implementation and just be presented with different images.

Mengmi:
> Do all three of these tasks use different mechanisms? I thought they would be based on the same "visual search" implementation and just be presented with different images. If it's the same mechanism, we should call it the same "visual search" task but then have three different benchmarks for this task.

Theoretically, they are the same; but because waldo and natural images are very large (1024x1280; downsampling is not possible since you cannot tell Waldo from the rest from low resolution images); we have to chop them into pieces to feed to the neural network and stitch output maps together. Thus, Shashi might have two different sets of input functions to process stimulis and post-processing. @shashikg  clarify pls if im wrong.

shashikg:
We can make similar implementation for the natural images and waldo cases but for the object array case, I guess it will differ, well everything can still be merged into a single program. But I guess @mschrimpf wants a general implementation that could work for any newly created stimuli if anyone wants to verify the benchmark on that new stimuli. @Mengmi I guess the main implementation which differs is inhibiting the complete object in object array case rather than a region around the fixation as in waldo and natural image cases. So, either we can change the name or keep a same name and add an additional attribute that will tell the benchmark to follow which kind of stimuli it is.

Mengmi:
let's use the same name and add in the attribute. Array is a special case. For any new stimuli set, we will always stick with the style to process Waldo/natural images.

mschrimpf:
Is there a way for the implementation to make the distinction based on image size then? If not, then let's just go with different tasks, it won't make a difference if we use one task with two attributes or just two different tasks.
If I'm being stupid and don't understand what you're saying, we can also quickly jump on a call to discuss this.

Mengmi:
It is not about image size. Since there are only 6 locations in object arrays, we extract fixations based on the binary mask within those 6 locations. We discard any fixations outside those 6 object locations. However, this is not the case for natural images and Waldo. This has nothing to do with image size. In natural images, we do not exclude any fixations since all fixations will land somewhere on the images. Yes, i'm fine with using two different tasks. One is for object arrays. period. The other task is for all the rest of images and all upcoming images in the future.

mschrimpf:
Sounds good, thanks @Mengmi and @shashikg!

mschrimpf:
@shashikg @Mengmi just wanted to check in how I can help to finalize this PR (and related ones). I think you're extremely close, only need to implement what we last talked about in this thread.

shashikg:
@mschrimpf  Hi Martin!
Sorry for the delay. I got a little busy with some of my projects. Thanks for reminding me. I will try to finish this in 1-2 weeks. As discussed before, I will implement the waldo and natural images as a different task, i.e. overall two task one for object cases and another one for general images.

shashikg:
@mschrimpf 

check this notebook file: https://github.com/shashikg/brain-score/blob/master/examples/visual-search-benchmark.ipynb

brainscore/benchmarks/search.py
```diff
@@ -85,3 +85,84 @@ def ceiling(self):
 
     def get_raw_data(self):
         return self.cumm_perf, self.saccades
+
+KlabZhang2018search_assembly = ['klab.Zhang2018search_waldo', 'klab.Zhang2018search_naturaldesign']
+KlabZhang2018search_assembly_ceil = [1, 1]
```
mschrimpf: is this the actual computed ceiling of the data? i.e. humans predict each other at 100%?

brainscore/benchmarks/search.py
```diff
@@ -85,3 +85,84 @@ def ceiling(self):
...
+KlabZhang2018search_assembly = ['klab.Zhang2018search_waldo', 'klab.Zhang2018search_naturaldesign']
+KlabZhang2018search_assembly_ceil = [1, 1]
+
+class KlabZhang2018VisualSearch(Benchmark):
+    def __init__(self, ceil_score=None, assembly_name=None):
+        if assembly_name is None:
+            self._logger.info("## No assembly name given, therefore selecting random assembly from 'waldo' or 'naturaldesign'...\n")
+            i = np.random.choice(2)
+            assembly_name = KlabZhang2018search_assembly[i]
+            ceil_score = KlabZhang2018search_assembly_ceil[i]
```
mschrimpf: Shouldn't we just have one benchmark for each of the two assemblies rather than selecting randomly?

brainscore/benchmarks/search.py
```diff
@@ -85,3 +85,84 @@ def ceiling(self):
...
+        self.data_len = int(self._assemblies.presentation.values.shape[0]/self.num_sub)
+        self.ior_size = 100
+
+        self._logger = logging.getLogger(fullname(self))
+
+    def __call__(self, candidate: BrainModel):
+        self._metric = ScanMatchPy.initialize()
+
+        self._logger.info("## Starting visual search task...")
+        candidate.start_task(BrainModel.Task.visual_search, max_fix=self.max_fix, data_len=self.data_len, ior_size=self.ior_size)
```
mschrimpf: the `data_len` is still a little unclear to me -- why does the model need to know about this?

brainscore/model_interface.py
```diff
@@ -12,7 +12,14 @@ class BrainModel:
     """
 
     RecordingTarget = Enum('RecordingTarget', " ".join(['V1', 'V2', 'V4', 'IT']))
-    Task = Enum('Task', " ".join(['passive', 'probabilities', 'label', 'visual_search_obj_arr']))
+    """
+    location to record from
+    """
+
+    Task = Enum('Task', " ".join(['passive', 'probabilities', 'label', 'visual_search_obj_arr', 'visual_search']))
```
mschrimpf: what are the differences between `visual_search_obj_arr` and `visual_search`? Can we document this somewhere?

brainscore/submission/ml_pool.py
```diff
@@ -24,10 +24,24 @@ def __contains__(self, item):
 
 
 class MLBrainPool(UniqueKeyDict):
-    def __init__(self, base_model_pool, model_layers):
+    def __init__(self, base_model_pool, model_layers, vs_model_param=None):
```
mschrimpf, line 27: What is `vs_model_param` for? Intuitively it seems like this parameter should not be here. If you want to add a specific new model you can implement it outside this generic `MLBrainPool`

brainscore/submission/ml_pool.py
```diff
@@ -24,10 +24,24 @@ def __contains__(self, item):
...
+            stimuli_model_pool = vs_model_param['stim_pool']
+            visual_search_layer = vs_model_param['model_layers']
+            target_img_size = vs_model_param['tar_size']
+            stimuli_img_size = vs_model_param['stim_size']
+        else:
+            target_model_pool = base_model_pool
+            stimuli_model_pool = base_model_pool
+            visual_search_layer = None
+            target_img_size = None
+            stimuli_img_size = None
```
mschrimpf, line 42: as above, I don't think this should be here

brainscore/submission/ml_pool.py
```diff
@@ -39,8 +53,28 @@ def __init__(self, base_model_pool, model_layers):
             def load(identifier=basemodel_identifier, activations_model=activations_model, layers=layers):
                 assert hasattr(activations_model, 'reload')
                 activations_model.reload()
+
+                search_target_model_param = {}
+                search_stimuli_model_param = {}
+                if (vs_model_param is not None) and (basemodel_identifier == 'vgg-16'): #as vs_layer is implemented only for vgg-16 as of now
```
mschrimpf: this should be a separate BrainModel then

brainscore/benchmarks/search.py
shashikg: seems you are looking at the outdated version. please check the latest one. i tagged you there.

brainscore/benchmarks/search.py
```diff
@@ -0,0 +1,168 @@
...
+        return self.human_score
+
+    def get_raw_data(self):
+        return self.cumm_perf, self.saccades
+    
+def KlabZhang2018VisualSearchWaldo():
+    return _KlabZhang2018VisualSearch(ceil_score=0.3519, assembly_name='klab.Zhang2018search_waldo', identifier_suffix="waldo")
+
+def KlabZhang2018VisualSearchNaturaldesign():
+    return _KlabZhang2018VisualSearch(ceil_score=0.3953, assembly_name='klab.Zhang2018search_naturaldesign', identifier_suffix="naturaldesign")
```
shashikg, line 168: @mschrimpf

brainscore/benchmarks/search.py
shashikg: again it's the outdated version. please check the latest one.

brainscore/benchmarks/search.py
shashikg: I will make the changes we can use stimulus_set for that. I added this parameter inside the benchmark init. So basically, we need data_len at some other places also. so instead of computing every time from the stimulus_set, I thought it may be much better to assign a variable to it. Note that the length of stimulus_set will be 3 times the data_len.

brainscore/model_interface.py
shashikg: again this is outdated. i renamed: 'object_search', 'visual_search'

recall the discussion with Mengmi. we have two benchmarks one benchmark for the experiment similar to the object array dataset. while the other one for waldo and natural design. 

object_search >> for object array
visual_search >> for other

Sure I will add the comments.

brainscore/submission/ml_pool.py
shashikg: This is what I was most confused about.

I guess you wanted me to do the implementation such that it can be aligned with this generic MLBrainPool. I can create a different one as well. In fact, I think that will be much better. Let me know what seems suitable to you.

PR #195, opened 2020-01-30
Remove deprecated hook code plus some submission fixes

franzigeiger:


submission/ml_pool.py
```diff
@@ -49,16 +50,6 @@ def __init__(self, base_model_pool, model_layers):
...
-                    brain_model = ModelCommitment(identifier=identifier, activations_model=activations_model,
-                                                  layers=layers)
-                    for region in regions:
-                        brain_model.commit_region(region)
-                    return brain_model
-
-                self[identifier] = LazyLoad(load)
+            self[basemodel_identifier] = LazyLoad(
+                lambda identifier=basemodel_identifier, activations_model=activations_model, layers=layers:
+                ModelCommitment(identifier=identifier, activations_model=activations_model, layers=layers))
```
mschrimpf: you're still referencing the `identifier` here, I think that should be `basemodel_identifier` -- or you rename `basemodel_identifier` to `identifier`

franzigeiger: The parameter is renamed in the lambda expression. `identifier=basemode_identifier`, so it is correct. This way I actually think it's more clear. One has to explicitly define lambda method parameters.

franzigeiger:
@mschrimpf The jenkins run fails, but that's not the fault of this PR. There are failing tests in brain-score which I can fix next. I would like this PR to merge though as it fixes errors happening in other projects.

PR #196, opened 2020-01-30
Tensorflow 2 update

franzigeiger:


mschrimpf:
Do the unit tests over in candidate_models still pass with these changes? https://github.com/brain-score/candidate_models/blob/bd545585ba5ccf16c80f8cb3a1f83ca874c6c8e3/tests/test___init__.py#L162-L173 If yes, then it looks good to me.
(would probably be preferable to move precomputed-feature versions of those tests into this repository)

PR #199, opened 2020-02-14
Screen degrees

tiagogmarques:
Added single neuron mapping
Added V1 benchmarks with single neuron mapping
Corrected bugs on the screen degrees transformation


brainscore/benchmarks/freemanziemba2013.py
```diff
@@ -40,13 +42,48 @@ def MovshonFreemanZiemba2013V1RDM():
...
+
+
+def MovshonFreemanZiemba2013V1PLSPublic():
+    return _MovshonFreemanZiemba2013RegionAccess('V1', identifier_metric_suffix='pls',
+                                                 similarity_metric=CrossRegressedCorrelation(
+                                                 regression=pls_regression(), correlation=pearsonr_correlation(),
+                                                 crossvalidation_kwargs=dict(stratification_coord='texture_type')),
+                                                 ceiler=InternalConsistency(), access='public')
+
+
```
mschrimpf: these should not be here. Public benchmarks are in https://github.com/brain-score/brain-score/blob/master/brainscore/public_benchmarks.py

brainscore/benchmarks/__init__.py
```diff
@@ -73,11 +73,15 @@ def __init__(self):
...
-            MovshonFreemanZiemba2013V1RDM, MovshonFreemanZiemba2013V2RDM
+            MovshonFreemanZiemba2013V1RDM, MovshonFreemanZiemba2013V2RDM, MovshonFreemanZiemba2013V1SINGLE, \
+            MovshonFreemanZiemba2013V1PLSPrivate, MovshonFreemanZiemba2013V1PLSPublic
         self['movshon.FreemanZiemba2013.V1-pls'] = LazyLoad(MovshonFreemanZiemba2013V1PLS)
         self['movshon.FreemanZiemba2013.V2-pls'] = LazyLoad(MovshonFreemanZiemba2013V2PLS)
         self['movshon.FreemanZiemba2013.V1-rdm'] = LazyLoad(MovshonFreemanZiemba2013V1RDM)
         self['movshon.FreemanZiemba2013.V2-rdm'] = LazyLoad(MovshonFreemanZiemba2013V2RDM)
+        self['movshon.FreemanZiemba2013.V1-single'] = LazyLoad(MovshonFreemanZiemba2013V1SINGLE)
+        self['movshon.FreemanZiemba2013.private.V1-pls'] = LazyLoad(MovshonFreemanZiemba2013V1PLSPrivate)
+        self['movshon.FreemanZiemba2013.public.V1-pls'] = LazyLoad(MovshonFreemanZiemba2013V1PLSPublic)
```
mschrimpf: delete, public benchmarks go to https://github.com/brain-score/brain-score/blob/master/brainscore/public_benchmarks.py

brainscore/metrics/regression.py
```diff
@@ -69,3 +90,25 @@ def linear_regression(xarray_kwargs=None):
...
+
+    xm = x - xmean
+    ym = y - ymean
+
+    normxm = scipy.linalg.norm(xm, axis=0, keepdims=True)
+    normym = scipy.linalg.norm(ym, axis=0, keepdims=True)
+
+    r = ((xm/normxm)*(ym/normym)).sum(axis=0)
+
+    return r
```
mschrimpf: is this different from `scipy.stats.pearsonr`?

brainscore/metrics/xarray_utils.py
```diff
@@ -88,7 +88,7 @@ def __call__(self, prediction, target):
             target_neuroids = target.isel(**{neuroid_dims[0]: i})  # `isel` is about 10x faster than `sel`
             prediction_neuroids = prediction.isel(**{neuroid_dims[0]: i})
             r, p = self._correlation(target_neuroids, prediction_neuroids)
-            correlations.append(r)
+            correlations.append(r if np.isfinite(r) else 0)
```
mschrimpf: let's just stick with `r`. If the correlation is `nan`, then that's what it is at this point, not 0.

brainscore/metrics/regression.py
tiagogmarques: It allows x to have multiple columns and gives the pearsonr columnwise x with y. Scipy only works with vectors. This makes it much faster for the single neuron regression mapping.

brainscore/metrics/xarray_utils.py
tiagogmarques: When one of the variable is constant, pearsonr gives nan. Mathematically, it should be zero as they are uncorrelated. This can happen a lot in the single neuron regression mapping when the units do not respond to the stimuli.

mschrimpf: I would really prefer to stick with the standard scipy output and only fiddle with it when needed, i.e. in this case in the single neuron regression (where we could drop `nan`s or set them to `0`).

PR #201, opened 2020-02-21
Little issues

franzigeiger:
I moved the submission module as you wanted it to be and implemented the model reloading as clean actual reload. There's no problem any more as every dict access loads a new instance of the model. It's a bit hard to see in the files because I also moved them(ml_pool abd utils)


brainscore/utils/__init__.py
```diff
@@ -67,6 +67,9 @@ def _ensure_loaded(self):
         if self.content is None:
             self.content = self.load_fnc()
 
+    def reload(self):
+        self.content = self.load_fnc()
```
mschrimpf: aah this is clever!

mschrimpf:
approved, but there seems to be a conflict

brainscore/submission/evaluation.py
```diff
@@ -55,8 +55,10 @@ def run_evaluation(config_file, work_dir, db_connection_config, jenkins_id, mode
         for model in test_models:
             function = lambda: module.get_model(model)
             base_model_pool[model] = LazyLoad(function)
-            if module.get_layers(model) is not None:
+            try:
                 layers[model] = module.get_layers(model)
+            except Exception:
+                pass
```
mschrimpf: Maybe add a `logging.warning(f'Could not retrieve layer for model {model} -- skipping model')` to make things more transparent

setup.py
```diff
@@ -20,6 +20,8 @@
     "pandas==0.25.3",
     "result_caching @ git+https://github.com/mschrimpf/result_caching",
     "jupyter",
+    "matplotlib",
```
mschrimpf: what's matplotlib used for?

setup.py
```diff
@@ -20,6 +20,8 @@
...
+    "tensorflow==1.15",
```
mschrimpf: so far, we left the modeling libraries out of the setup.py so that people can set them up with conda, but this is probably for the OST stuff right?

tests/test_benchmarks/test___init__.py
```diff
@@ -107,13 +107,21 @@ def run_test(self, benchmark, file, expected):
...
             precomputed_features[column] = 'presentation', stimulus_set[column].values
         precomputed_features = PrecomputedFeatures(precomputed_features)
         # score
         score = benchmark(precomputed_features).raw
         assert score.sel(aggregation='center') == expected
+
+
+def lstrip_local(path):
+    parts = path.split(os.sep)
+    path = os.sep.join(parts[-3:])
```
mschrimpf: I think it would be better to keep this flexible with respect to the root directory (i.e. like model-tools is doing it for `~/.brainio` https://github.com/brain-score/model-tools/blob/b378404216bde499562f602670d929062b357605/model_tools/activations/core.py#L224). I'd just use the same function here.

setup.py
franzigeiger: Jupyter notebook tests

setup.py
franzigeiger: Yes but this is a problem for the CI. If the library is not there it will fail, so from now on to pass the test gate on jenkins you will have to add all requried libraries in setup.py. Otherwise they are not installed. And yes I need tensorflow for the OST stuff afaik.

tests/test_benchmarks/test___init__.py
franzigeiger: I tried that but it's a bit messy. The paths can be different after `/brainio` sometimes images are in another `data` folder and sometimes they are not. I could not really figure out which version is the actual correct one but either way I want the test to work if the correct images are available, so I rewrote the test this way. Additionally if I would use the model tools implementation I would have to add model tools as a dependency to brain-score which is not the case right now and I would like to leave it like that.

franzigeiger:
The CI environment fails because stupid jenkins gives me a deprecated git commit hash. https://issues.jenkins-ci.org/browse/JENKINS-56854 I have not yet figured out why exactly that is and would like to have a green pass, but for now I'm almost confident that all tests will pass and merge this PR.

PR #202, opened 2020-02-27
Submission check module

franzigeiger:
Move submission check module to brain-score


franzigeiger:
Moved the module to model-tools

PR #204, opened 2020-04-09
use logistic OSTs

mschrimpf:
@kohitij-kar this should be your call. CORnet-S scores seem to improve quite a bit when comparing based on logistic OSTs.


mschrimpf:
waiting for @kohitij-kar to recompute ceiling with new python pipeline

mschrimpf:
closing due to inactivity

PR #207, opened 2020-04-29
fix Rajalingham2018public benchmark ceiling and verify with unit test

mschrimpf:
The public data does not have a full set of trials for all image/distractor pairs and as a result, there are some nans in the data which didn't let us compute the ceiling. We already had code to deal with it, it was just a matter of turning on the necessary flag to ignore nans in the split-half source of the metric.

### DIFF
--- a/brainscore/benchmarks/public_benchmarks.py
+++ b/brainscore/benchmarks/public_benchmarks.py
@@ -64,6 +64,7 @@ class RajalinghamMatchtosamplePublicBenchmark(DicarloRajalingham2018I2n):
     def __init__(self):
         super(RajalinghamMatchtosamplePublicBenchmark, self).__init__()
         self._assembly = LazyLoad(lambda: load_rajalingham2018(access='public'))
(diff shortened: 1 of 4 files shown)


mschrimpf:
jenkins is unfortunately still failing since it isn't downloading the required test files (https://github.com/brain-score/brain-score/blob/master/tests/download_test_files.sh)

franzigeiger:
So the script has to be run manually? Can you call it in a before method of the unit from python code?

mschrimpf:
for travis, I just specified to execute the script before the tests: https://github.com/brain-score/brain-score/blob/98d5f3893b84398e1744dbfb05dfe8e4b37831d1/.travis.yml#L18
We could manually do the same in jenkins, but as we talked about it would be ideal if we could somehow reuse the `.yml` to avoid duplicate efforts (and putting things in a non-versioned jenkins config is much less transparent).

mschrimpf:
to be clear for this PR, we could opt to only worry about jenkins after the NeurIPS deadline and merge it regardless

franzigeiger:
Ok fair!

PR #216, opened 2020-05-07
return similarity Score instead of DataAssembly; allow nans in diagonal

mschrimpf:


mschrimpf:
travis tests pass; only tests with missing precomputed assemblies fail on jenkins

PR #223, opened 2020-07-16
Secret manager + Database schema

franzigeiger:
- Adjusted code for proper secret manager usage 
- Use new database scheme
- Remove initial layer commitment


brainscore/submission/database.py
```diff
@@ -4,26 +4,23 @@
...
 
-def store_score(database_secret, score):
-    dbConnection = connect_db(database_secret)
+def store_score(db_configs, score):
+    dbConnection = connect_db(db_configs)
     try:
         insert = '''insert into benchmarks_score
-                (model, benchmark, score_raw, score_ceiled, error, timestamp, jenkins_job_id, user_id, name)   
-                VALUES(%s,%s,%s,%s,%s,%s, %s, %s, %s)'''
+                (model, benchmark, score_raw, score_ceiled, error, layer, timestamp, jenkins_job_id, owner, author)
```
mschrimpf, line 17: there is no `layer` column anymore (https://github.com/brain-score/brain-score.web/pull/15/files#diff-fd9b76181a534a55f1ed5ae657dcafe0L121), so this will fail

franzigeiger: No, it actually won't fail, because the production database still has the old schema. This PR is a bit older and I forgot to merge it. I need this merged though because the next PR  #224 is based on this one.

PR #225, opened 2020-07-28
Rework submission system

franzigeiger:
- Add option to rerun models
- Use new database scheme
- Extend submission 'API' by a new method `get_bibtex(model_identifier)`

Still to do:
- [x] Write tests (it is time to do that)


brainscore/benchmarks/imagenet.py
```diff
@@ -21,7 +21,16 @@ def __init__(self):
         super(Imagenet2012, self).__init__(identifier='fei-fei.Deng2009-top1', version=1,
                                            ceiling_func=lambda: ceiling,
                                            parent='ImageNet',
-                                           paper_link="https://ieeexplore.ieee.org/abstract/document/5206848")
+                                           paper_link="https://ieeexplore.ieee.org/abstract/document/5206848",
```
mschrimpf, line 24: as above, paper_link should become url in bibtex

brainscore/benchmarks/kar2019.py
```diff
@@ -16,7 +16,26 @@ def __init__(self):
         super(DicarloKar2019OST, self).__init__(identifier='dicarlo.Kar2019-ost', version=2,
                                                 ceiling_func=lambda: ceiling,
                                                 parent='IT-temporal',
-                                                paper_link='https://www.nature.com/articles/s41593-019-0392-5')
+                                                paper_link='https://www.nature.com/articles/s41593-019-0392-5',
```
mschrimpf, line 19: as above, paper_link should become url in bibtex

brainscore/benchmarks/majajhong2015.py
```diff
@@ -15,7 +15,22 @@ def _DicarloMajajHong2015Region(region, identifier_metric_suffix, similarity_met
     return NeuralBenchmark(identifier=f'dicarlo.MajajHong2015.{region}-{identifier_metric_suffix}', version=3,
                            assembly=assembly, similarity_metric=similarity_metric, visual_degrees=VISUAL_DEGREES,
                            ceiling_func=lambda: ceiler(assembly_repetition),
-                           parent=region, paper_link='http://www.jneurosci.org/content/35/39/13402.short')
+                           parent=region, paper_link='http://www.jneurosci.org/content/35/39/13402.short', bibtex =
```
mschrimpf, line 18: as above, paper_link should become url in bibtex

brainscore/benchmarks/public_benchmarks.py
```diff
@@ -37,7 +37,22 @@ def _standard_benchmark(identifier, load_assembly, visual_degrees, stratificatio
     return NeuralBenchmark(identifier=f"{identifier}-pls", version=1,
                            assembly=assembly, similarity_metric=similarity_metric, visual_degrees=visual_degrees,
                            ceiling_func=lambda: ceiler(assembly_repetition),
-                           parent=None, paper_link='http://www.jneurosci.org/content/35/39/13402.short')
+                           parent=None, paper_link='http://www.jneurosci.org/content/35/39/13402.short',
```
mschrimpf, line 40: as above, paper_link should become url in bibtex

brainscore/benchmarks/public_benchmarks.py
```diff
@@ -37,7 +37,22 @@ def _standard_benchmark(identifier, load_assembly, visual_degrees, stratificatio
...
+                           bibtex= """@article {Majaj13402,
```
mschrimpf, line 41: use different bibtex for different public benchmarks, the way it's written now would use the Majaj&Hong bibtex also for Freeman&Ziemba. I suggest treating those like e.g. the `visual_degrees` and specifying them in the respective benchmark files as constants and then importing.

brainscore/benchmarks/rajalingham2018.py
```diff
@@ -20,7 +20,19 @@ def __init__(self):
             identifier='dicarlo.Rajalingham2018-i2n', version=2,
             ceiling_func=lambda: self._metric.ceiling(self._assembly),
             parent='behavior',
-            paper_link='https://www.biorxiv.org/content/early/2018/02/12/240614')
+            paper_link='https://www.biorxiv.org/content/early/2018/02/12/240614',
```
mschrimpf, line 23: as above, paper_link should become url in bibtex

brainscore/benchmarks/__init__.py
```diff
@@ -73,12 +73,13 @@ class BenchmarkBase(Benchmark):
...
     Helper class for implementing standard functions of the `Benchmark` interface.
     """
-    def __init__(self, identifier, ceiling_func, version, parent=None, paper_link=None):
+    def __init__(self, identifier, ceiling_func, version, parent=None, paper_link=None, bibtex=None):
         self._identifier = identifier
         self._ceiling_func = ceiling_func
         self._version = version
         self.parent = parent
         self.paper_link = paper_link
+        self.bibtex = bibtex
```
mschrimpf, line 82: maybe we should also add this as a property to the `Benchmark` interface (which `BenchmarkBase` implements), i.e.:
```
    @property
    def bibtex(self):
        """
        bibtex string to build the reference.
        Should include an `url` to build a proper link.
        """
        raise NotImplementedError()
```

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+from brainscore.submission.models import Model, Submission
+
+
+class BaseConfig:
+    def __init__(self, work_dir, jenkins_id, db_secret, config_path):
+        self.work_dir = work_dir
+        self.jenkins_id = jenkins_id
+        self.config_path = config_path
+        self.db_secret = db_secret
+        return
```
mschrimpf, line 13: unnecessary code

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+class MultiConfig(BaseConfig):
+    def __init__(self, model_ids, **kwargs):
+        BaseConfig.__init__(self, **kwargs)
+        self.models = []
+        self.submissions = {}
+        for id in model_ids:
+            model = Model.get(id=id)
+            submission: Submission = model.submission
```
mschrimpf, line 23: ooh fancy, didn't know about this syntax (have only used it in method parameters so far)

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+            self.models.append(model)
+            if submission.id not in self.submissions:
+                self.submissions[submission.id] = submission
+
+
+class SubmissionConfig(BaseConfig):
+    def __init__(self, model_type, user_id, jenkins_id, public, **kwargs):
+        BaseConfig.__init__(self, jenkins_id=jenkins_id, **kwargs)
```
mschrimpf, line 31: as above with the super call

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+        self.work_dir = work_dir
+        self.jenkins_id = jenkins_id
+        self.config_path = config_path
+        self.db_secret = db_secret
+        return
+
+
+class MultiConfig(BaseConfig):
+    def __init__(self, model_ids, **kwargs):
+        BaseConfig.__init__(self, **kwargs)
```
mschrimpf, line 18: is there a reason you're not using the `super` call? It doesn't matter in this case (yet), but if there is ever multiple inheritance, `super` guarantees calling every method exactly once (see e.g. https://stackoverflow.com/a/21639994/2225200)

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+class BaseConfig:
+    def __init__(self, work_dir, jenkins_id, db_secret, config_path):
+        self.work_dir = work_dir
+        self.jenkins_id = jenkins_id
+        self.config_path = config_path
+        self.db_secret = db_secret
+        return
+
+
+class MultiConfig(BaseConfig):
```
mschrimpf, line 16: Some documentation/comments here would be helpful to understand what the difference between `MultiConfig` and `SubmissionConfig` is and when to use which one

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def object_decoder(config, work_dir, config_path, db_secret, jenkins_id):
+    if 'model_ids' in config:
+        return MultiConfig(model_ids=config['model_ids'], work_dir=work_dir, config_path=config_path,
+                           jenkins_id=jenkins_id, db_secret=db_secret)
+    else:
+        return SubmissionConfig(model_type=config['model_type'], user_id=config['user_id'], work_dir=work_dir,
+                                config_path=config_path,
+                                jenkins_id=jenkins_id, db_secret=db_secret, public=bool(distutils.util.strtobool(config['public'])))
```
mschrimpf, line 44: recommend simplifying to `public=config['public']==True`

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,44 @@
...
+
+class SubmissionConfig(BaseConfig):
+    def __init__(self, model_type, user_id, jenkins_id, public, **kwargs):
+        BaseConfig.__init__(self, jenkins_id=jenkins_id, **kwargs)
+        self.submission = Submission.create(id=jenkins_id, submitter=user_id, timestamp=datetime.datetime.now(),
+                                            model_type=model_type, status='running')
+        self.public = public
+
+
+def object_decoder(config, work_dir, config_path, db_secret, jenkins_id):
```
mschrimpf, line 37: also a bit confused about this method, a short comment would probably clear that up

brainscore/submission/developers_guide.md
```diff
@@ -0,0 +1,53 @@
+## Submission system
```
mschrimpf, line 1: Awesome!!!

brainscore/submission/evaluation.py
```diff
@@ -24,47 +22,64 @@
                        if benchmark not in ['dicarlo.Kar2019-ost', 'fei-fei.Deng2009-top1']]
```
mschrimpf, line 22: we should figure out a way soon to run these efficiently. Maybe execution in parallel would help so that we don't have to wait for every single benchmark to finish?

brainscore/submission/evaluation.py
```diff
@@ -79,21 +94,25 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
...
+                        model.layer_model.region_layer_map) if submission.model_type == 'BaseModel' else ''
                     result = {
                         'Model': model_id,
                         'Benchmark': benchmark,
                         'raw_result': raw,
                         'ceiled_result': ceiled,
                         'error': error,
                         'finished_time': finished,
-                        'layer' : str(model.layer_model.region_layer_map)
+                        'layer': layer_commitment
```
mschrimpf, line 106: I wrote this in the other PR, but I think this should be something like `'comment': "layers: {layer_commitment}"`

brainscore/submission/evaluation.py
```diff
@@ -24,47 +22,64 @@
...
     try:
-        for model_id in test_models:
+        for modelInst in test_models:
+            model_id = modelInst.name
             for benchmark in test_benchmarks:
+                scoreInst = None
                 try:
+                    start = datetime.datetime.now()
+                    bench_inst = get_benchmark_instance(benchmark)
+                    scoreInst = Score.create(benchmark=bench_inst, start_timestamp=start, model=modelInst)
```
mschrimpf, line 82: there's some inconsistent naming here (`bench_inst` vs `scoreInst`), would prefer `benchmark_instance` and `score`

brainscore/submission/evaluation.py
```diff
@@ -102,48 +121,69 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
                         'raw_result': 0, 'ceiled_result': 0,
                         'error': error, 'finished_time': datetime.datetime.now()
                     })
+                    if scoreInst:
+                        scoreInst.comment = error
```
mschrimpf, line 125: following the change with layers, this might then have to be `comment = error if comment is None else f"{comment}\n{error}` to append to the comment?

brainscore/submission/evaluation.py
```diff
@@ -102,48 +121,69 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
...
+        ml_brain_pool = MLBrainPool(base_model_pool, model_layers)
+    else:
+        logger.info(f"Start working with brain models")
+        for model in test_models:
+            ml_brain_pool[model.name] = module.get_model(model.name)
+    return ml_brain_pool
+
+
+def get_benchmark_instance(benchmark):
+    bench = benchmark_pool[benchmark]
```
mschrimpf, line 161: might as well write it out to `benchmark` :)

brainscore/submission/evaluation.py
```diff
@@ -102,48 +121,69 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
...
+    benchmark_type, created = BenchmarkType.get_or_create(identifier=benchmark, order=999)
+    if created:
+        try:
+            parent = BenchmarkType.get(identifier=bench.parent)
+            benchmark_type.parent = parent
+            benchmark_type.save()
+        except DoesNotExist:
+            logger.error(
+                f'Couldn\'t connect benchmark {benchmark} to parent {bench.parent} since parent doesn\'t exist')
```
mschrimpf, line 170: I would either raise an error here or turn this into a warning

brainscore/submission/evaluation.py
```diff
@@ -102,48 +121,69 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
...
+        bench_inst.ceiling = ceiling.sel(aggregation='center')
+        bench_inst.ceiling_error = ceiling.sel(aggregation='error')
+        bench_inst.save()
+    return bench_inst
+
+
+def get_reference(bibtex_string):
+    parsed = bibtexparser.loads(bibtex_string)
+    entry = list(parsed.entries)[0]
+    ref, create= Reference.get_or_create(bibtex=bibtex_string, author=entry['author'], url=entry['url'], year=entry['year'])
```
mschrimpf, line 188: bit pedantic but: autoformat, there's a missing space before the `=`

mschrimpf: also does this deal with the case where a reference does not provide the url? (or do we force every benchmark to have that? then we should add a small unit test to check for url existence in each benchmark's bibtex)

brainscore/submission/repository.py
```diff
@@ -0,0 +1,57 @@
...
+    list = os.listdir(work_dir)
+    candidates = []
+    for item in list:
+        if not item.startswith('.') and not item.startswith('_'):
+            candidates.append(item)
+    if len(candidates) is 1:
+        return candidates[0]
+    logger.error('The zip file structure is not correct, we try to detect the correct directory')
+    if 'sample-model-submission' in candidates:
+        return 'sample-model-submission'
```
mschrimpf, line 45: is this for testing only? Can you add an explanatory comment?

brainscore/submission/repository.py
```diff
@@ -0,0 +1,57 @@
...
+    logger.error('The submission file contains too many entries and can therefore not be installed')
+    raise Exception('The submission file contains too many entries and can therefore not be installed')
+
+
+def install_project(repo, package):
+    try:
+        assert 0 == subprocess.call([sys.executable, "-m", "pip", "install", "-v", repo], env=os.environ)
```
mschrimpf, line 52: maybe add an error message that the installation failed in case this goes wrong?

tests/test_submission/base_model.py
```diff
@@ -0,0 +1,50 @@
...
+import torchvision.models
+from model_tools.activations.pytorch import PytorchWrapper
+from model_tools.activations.pytorch import load_preprocess_images
+
+# This is an example implementation for submitting alexnet as a pytorch model
+# If you use pytorch, don't forget to add it to the setup.py
+
+# Attention: It is important, that the wrapper identifier is unique per model!
+# The results will otherwise be the same due to brain-scores internal result caching mechanism.
+# Please load your pytorch model for usage in CPU. There won't be GPU's available for scoring your model.
```
mschrimpf, line 12: GPU's -> GPUs

tests/test_submission/test_integration.py
```diff
@@ -0,0 +1,106 @@
...
+from brainscore.submission.database import connect_db
+from brainscore.submission.evaluation import run_evaluation
+from brainscore.submission.models import Score, Model, Submission
+from tests.test_submission.test_db import clear_schema, init_user
+
+logger = logging.getLogger(__name__)
+
+#
+#     Integration tests for the submission systems, executing 4 submissions:
+#     1: ID:33 Working submission, executing one benchmark on Alexent (zip + json)
```
mschrimpf, line 15: Alexent -> Alexnet

tests/test_submission/test_integration.py
```diff
@@ -0,0 +1,106 @@
...
+#     6: ID:36 Submission is installable, but model (Alexnet) is not scoreable (zip + json)
+#
+
+
+class TestIntegration:
+    databse = 'brainscore-ohio-test'
+    @classmethod
+    def setup_class(cls):
+        logger.info('Connect to database')
+        # connect_db('brainscore-ohio-test')
```
mschrimpf, line 27: dead code

tests/test_submission/test_submission.py
```diff
@@ -0,0 +1,185 @@
...
+from brainscore.submission.evaluation import get_reference, get_benchmark_instance, get_ml_pool, \
+    run_submission
+from brainscore.submission.models import Reference, BenchmarkType, Submission, Model, BenchmarkInstance, Score
+from brainscore.submission.repository import prepare_module, extract_zip_file, find_correct_dir
+from model_tools.brain_transformation import ModelCommitment
+from tests.test_submission import base_model
+from tests.test_submission.test_db import clear_schema, init_user
+
+logger = logging.getLogger(__name__)
+# database = 'brainscore-1-ohio-cred'
```
mschrimpf, line 19: dead code

tests/test_submission/test_submission.py
```diff
@@ -0,0 +1,185 @@
...
+database = 'brainscore-ohio-test'
+
+class TestSubmission:
+
+    @classmethod
+    def setup_class(cls):
+        logger.info('Connect to database')
+        # connect_db()
```
mschrimpf, line 27: dead code

brainscore/benchmarks/majajhong2015.py
```diff
@@ -15,7 +15,22 @@ def _DicarloMajajHong2015Region(region, identifier_metric_suffix, similarity_met
     return NeuralBenchmark(identifier=f'dicarlo.MajajHong2015.{region}-{identifier_metric_suffix}', version=3,
                            assembly=assembly, similarity_metric=similarity_metric, visual_degrees=VISUAL_DEGREES,
                            ceiling_func=lambda: ceiler(assembly_repetition),
-                           parent=region, paper_link='http://www.jneurosci.org/content/35/39/13402.short')
```
mschrimpf, line 18: I think you inadvertently removed the parent here too

brainscore/benchmarks/public_benchmarks.py
```diff
@@ -37,7 +37,21 @@ def _standard_benchmark(identifier, load_assembly, visual_degrees, stratificatio
     return NeuralBenchmark(identifier=f"{identifier}-pls", version=1,
                            assembly=assembly, similarity_metric=similarity_metric, visual_degrees=visual_degrees,
                            ceiling_func=lambda: ceiler(assembly_repetition),
-                           parent=None, paper_link='http://www.jneurosci.org/content/35/39/13402.short')
```
mschrimpf, line 40: I think you inadvertently removed the `parent` parameter here

brainscore/submission/evaluation.py
franzigeiger: I could run models in parallel, but I wouldn't do the same with benchmarks. After all, they don't take too long once all data is loaded.

brainscore/submission/evaluation.py
franzigeiger: ok I tried to make it more transparent. Database objects are called `_entry`, strings are called `_name` and brainscore objects keep the base name

brainscore/submission/evaluation.py
franzigeiger: NO because, this comment is stored in the database, the other comment is only sent to the submitter in the csv file. We are not storing the layer commitment in the database

brainscore/submission/evaluation.py
franzigeiger: I am checking that beforehand and only call this if the method is available

brainscore/submission/repository.py
franzigeiger: No this is just trying to guess the correct file in case there are multiple. Has nothing to do with tests, I just thought in case sb didn't rename the folder we can detect it.

brainscore/submission/evaluation.py
mschrimpf: aah gotcha, I thought the other comment was also for the database.
Models now have a comment field though, so could we store the layer commitment there?

brainscore/submission/configuration.py
```diff
@@ -0,0 +1,63 @@
...
+    This should help to better understand which properties are available and reformats properly.
+    """
+    if 'model_ids' in config:
+        return MultiConfig(model_ids=config['model_ids'], work_dir=work_dir, config_path=config_path,
+                           jenkins_id=jenkins_id, db_secret=db_secret)
+    else:
+        return SubmissionConfig(model_type=config['model_type'], user_id=config['user_id'], work_dir=work_dir,
+                                config_path=config_path,
+                                jenkins_id=jenkins_id, db_secret=db_secret,
+                                public=config['public'] == True)
```
mschrimpf, line 63: I missed the quotation marks in my previous comment, I believe it should be `== "True"`

brainscore/submission/evaluation.py
mschrimpf: ImageNet takes a long time IIRC, that's mostly why we're excluding it (and it requires GPU). The number of benchmarks will also increase more and more. (no need to address this in this PR but maybe something we should discuss)

brainscore/submission/evaluation.py
```diff
@@ -82,70 +116,97 @@ def run_evaluation(config_file, work_dir, jenkins_id, db_secret, models=None,
...
                     result = {
                         'Model': model_id,
-                        'Benchmark': benchmark,
+                        'Benchmark': benchmark_name,
                         'raw_result': raw,
                         'ceiled_result': ceiled,
                         'error': error,
                         'finished_time': finished,
-                        'layer' : str(model.layer_model.region_layer_map)
+                        'comment': {'layers': layer_commitment}
```
mschrimpf, line 128: just to make sure I get this right now: this comment will only be in the csv sent to the submitter and not show up in the database? (either way, shouldn't the `{'layers': layer_commitment}` just be a string `f"layers: {layer_commitment}"`?)

brainscore/submission/repository.py
mschrimpf: ok can you please add a short comment saying exactly that?

brainscore/submission/evaluation.py
franzigeiger: Yes it is only in the csv, I made it a string!

mschrimpf:
ok approved. Looks like there are a couple merge conflicts, but other than that things are good to merge

PR #227, opened 2020-07-31
implement looking at multiple trials to enable stochastic models

mschrimpf:
- [x] repeat stimulus_set with fixed number of repetitions
- [ ] ~repeat stimulus_set with per-image repetitions~
- [x] average model per-trial responses
- [x] update current benchmarks to use trials
- [x] ensure unit-tests pass
- [x] adjust model-tools for efficient trials
- [x] adjust model-tools for ImageNet assembly

### DIFF
--- a/brainscore/benchmarks/_neural_common.py
+++ b/brainscore/benchmarks/_neural_common.py
@@ -3,12 +3,12 @@ import numpy as np
 from brainio_base.assemblies import array_is_element, walk_coords
 from brainscore.benchmarks import BenchmarkBase, ceil_score
 from brainscore.benchmarks.screen import place_on_screen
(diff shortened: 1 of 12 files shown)


tiagogmarques:
Majaj: ~50 repetitions
Ziemba: 20 repetitions
Rajalingham: ~2 repetitions
Cadena: “at least 2 repetitions” (if I remember in Brain-score you remove images that have NaNs so I think it leaves out only the images that had 4 repetitions but it may be worth revisiting
Kar: 44 repetitions

mschrimpf:
using 10 trials for ImageNet since that is what many model ensembles use

mschrimpf:
@tiagogmarques approved

PR #232, opened 2020-09-14
adding ImageNet_C benchmark

dapello:
I think it should look something like this -- I'm not so sure about the update to the `__init__.py` file; maybe I should be calling the noise categories individually?

Also, I haven't editted Accuracy yet. Still think that needs a tad more thought.

Take a look and let me know what you all think.


dapello:
I've made the changes we discussed, feel free to review now.

brainscore/benchmarks/imagenet_c.py
```diff
@@ -0,0 +1,145 @@
...
+
+import brainscore
+from brainscore.benchmarks import BenchmarkBase
+from brainscore.benchmarks.trials import repeat_trials, average_trials
+from brainscore.metrics import Score
+from brainscore.metrics.accuracy import Accuracy
+from brainscore.model_interface import BrainModel
+from brainio_base.stimuli import StimulusSet
+
+BIBTEX="""@ARTICLE{Hendrycks2019-di,
```
mschrimpf, line 15: is there a url? the bibtex's `url` field is what we use to link to the paper from the website

tests/test_benchmarks/test_imagenet_c.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+import pytest
+from pytest import approx
+
+from brainio_base.assemblies import BehavioralAssembly
+from brainscore.benchmarks.imagenet import Imagenet2012
+from brainscore.benchmarks.imagenet_c import Imagenet_C_Individual, Imagenet_C_Noise, Imagenet_C_Blur, Imagenet_C_Weather, Imagenet_C_Digital
+from brainscore.model_interface import BrainModel
+
+#@pytest.mark.slow
```
mschrimpf, line 10: this is commented out -- following slack something breaks if you comment it in, but it's still not clear to me what exactly goes wrong

mschrimpf (approved):

mschrimpf:
ok the failing unit tests are:
* `TestPoolList.test_exact_evaluation_pool` because the ImageNetC benchmarks are not in the evaluation pool
* `test_imagenet_c.py` because `brainio_collection.lookup.StimulusSetLookupError: 'stimulus_set dietterich.Hendrycks2019.gaussian_noise_1 not found'`, i.e. it's not packaged?

mschrimpf:
re-triggering tests after https://github.com/brain-score/brainio_collection/pull/41 was merged

tests/test_benchmarks/test_imagenet_c.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+import pytest
+from pytest import approx
+
+from brainio_base.assemblies import BehavioralAssembly
+from brainscore.benchmarks.imagenet import Imagenet2012
+from brainscore.benchmarks.imagenet_c import Imagenet_C_Individual, Imagenet_C_Noise, Imagenet_C_Blur, Imagenet_C_Weather, Imagenet_C_Digital
+from brainscore.model_interface import BrainModel
+
```
mschrimpf, line 9: the tests in this file fail because the stimulus_sets are currently private and the test is not marked as `pytest.mark.private_access` (see [here](https://github.com/brain-score/brain-score/blob/master/tests/README.md#markers)).
You can either tell me to make the stimulus_sets public (if we're allowed to distribute them publicly) or add the mark

mschrimpf:
updated S3 policy: make only .zip public (including .csv goes above the max policy size)

mschrimpf:
@dapello when you re-packaged things, your lookup.csv was updated yes? for some reason the SHA1 hashes are different (this happened to me too, I'm guessing it's because zipping is non-deterministic) -- can you please update/renew a PR in brainio_collection to update the lookup.csv with the correct hashes? that's the only reason for the failing test as far as I can tell.

mschrimpf:
re-trigger with updated lookup.csv

PR #235, opened 2020-09-22
have model deal with averaging over trials

mschrimpf:
The current solution to presenting multiple trials to models and averaging over them repeats stimuli and then averages model activations. This unfortunately blows up memory requirements and as a short-term solution to this technical obstacle, this PR offloads the averaging over trials onto the model (where the non-stochastic models can ignore the parameter, and stochastic ones can e.g. use a running average).


tiagogmarques (approved):

mschrimpf:
the only failing test is imagenet which is solved by #232 -- merging

PR #240, opened 2020-09-23
add metadata (parent, bibtex, number of trials) to Cadena benchmark

mschrimpf:


mschrimpf:
only ImageNet test is failing which is fixed through #232

PR #242, opened 2020-09-30
remove ImageNet-C stimuli identifier to disable caching activations

mschrimpf:
* the activations are not used a second time, like for the ImageNet benchmark. so we would only store unnecessary information
* caching for temporal models like CORnet-S breaks when checking that coordinates are present because `logits-t0` doesn't match `logits`


mschrimpf:
test failures primarily from S3 outage

PR #246, opened 2020-11-02
add CKA metric

mschrimpf:
implementation from https://github.com/yuanli2333/CKA-Centered-Kernel-Alignment


mschrimpf:
re-trigger jenkins

PR #251, opened 2020-12-15
Update pandas version

jjpr-mit:
Update to the latest version of pandas.


mschrimpf:
re-trigger tests

mschrimpf:
I'm gonna merge because only the submission tests fail which should be unrelated to this, but we should still fix those soon.

PR #253, opened 2021-01-29
Tying the build to Python 3.6 caused problems with newer libraries.

jjpr-mit:


mschrimpf:
re-triggering jenkins

mschrimpf:
re-triggering jenkins^2

mschrimpf (approved):

PR #259, opened 2021-03-15
if candidate recordings are not temporal, exit early in Kar2019

mschrimpf:
evaluating models on Kar2019 takes notoriously long and often fails with out-of-memory errors (see e.g. http://braintree.mit.edu:8080/job/run_benchmarks/813 for a recent example) -- but ultimately most models (feedforward) get a nan score (cannot predict) anyway.

The out-of-memory errors occur especially for models with large activations when repeating their activations over time (https://github.com/brain-score/model-tools/blob/87e0a0fb5aec46e6c2c4aae06a3eb2040752e52c/model_tools/brain_transformation/temporal.py#L39-L47).

To solve the unnecessary use of compute resources for feedforward models, this PR runs a single image on the model candidate and checks whether any of the neuroids vary over time. If not, it exits early with a nan score.

### DIFF
--- a/brainscore/benchmarks/kar2019.py
+++ b/brainscore/benchmarks/kar2019.py
@@ -7,7 +7,28 @@ from brainscore.metrics import Score
 from brainscore.metrics.ost import OSTCorrelation
 from brainscore.model_interface import BrainModel
 
+BIBTEX = """@Article{Kar2019,
+            author={Kar, Kohitij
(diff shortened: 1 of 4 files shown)


stothe2 (approved):

mschrimpf:
jenkins test timed out, but the Kar2019 tests passed. Merging...

PR #265, opened 2021-03-30
Tissue Correlation Benchmark draft PR

pmcgrath249:


brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+class DicarloLee2020ITTissueResponseCorrelation(BenchmarkBase):
+
+    def __init__(self, metric=ks_similarity, bootstrap_samples=100_000, num_sample_arrays_candidate=10, ceiler_fn=InternalConsistency):
+
+        def load_target_assembly():
+            assembly = make_static(brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT'))
+            return assembly[:, np.in1d(assembly.presentation.image_id, list(set(stimulus_set['image_id'])))]
+
+
+        stimulus_set = Rajalingham2019()._target_assembly.stimulus_set
```
mschrimpf, line 34: I think this should be moved down one line and be `stimulus_set = assembly.stimulus_set`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+            URL = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116},
+            eprint = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116.full.pdf},
+            journal = {bioRxiv}}'''
+
+class DicarloLee2020ITTissueResponseCorrelation(BenchmarkBase):
+
+    def __init__(self, metric=ks_similarity, bootstrap_samples=100_000, num_sample_arrays_candidate=10, ceiler_fn=InternalConsistency):
+
+        def load_target_assembly():
+            assembly = make_static(brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT'))
```
mschrimpf, line 30: separate into two lines to reduce clutter:
``` 
assembly = brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT')
assembly = make_static(assembly)
```

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+            return assembly[:, np.in1d(assembly.presentation.image_id, list(set(stimulus_set['image_id'])))]
+
```
mschrimpf, line 32: actually I don't think this needs to be a function, you call it anyway in line 35. Instead put it directly as code inside the `__init__`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+        self._stimulus_set = stimulus_set
+        self._target_assembly = average_repetition(assembly)
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
+        self._metric = metric
+
+        self.bootstrap_samples = bootstrap_samples
+        self.num_sample_arrs = num_sample_arrays_candidate
+        self._arr_size_mm = (np.ptp(self._target_assembly.neuroid.x.data),
+                             np.ptp(self._target_assembly.neuroid.y.data))
```
mschrimpf, line 51: comment what this variable is, I didn't immediately understand what it contains

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+            URL = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116},
+            eprint = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116.full.pdf},
+            journal = {bioRxiv}}'''
+
+class DicarloLee2020ITTissueResponseCorrelation(BenchmarkBase):
+
+    def __init__(self, metric=ks_similarity, bootstrap_samples=100_000, num_sample_arrays_candidate=10, ceiler_fn=InternalConsistency):
```
mschrimpf, line 27: remove `ceiler_fn` from the `__init__`, this should be fixed. Ideally the other variables should also not be defined in the `__init__`. If you still want to change them for development purposes, you can always set `benchmark._bootstrap_samples = 123`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
+        self._metric = metric
+
+        self.bootstrap_samples = bootstrap_samples
+        self.num_sample_arrs = num_sample_arrays_candidate
+        self._arr_size_mm = (np.ptp(self._target_assembly.neuroid.x.data),
+                             np.ptp(self._target_assembly.neuroid.y.data))
+
+    def __call__(self, candidate):
+
```
mschrimpf, line 54: remove newline to comply with code format in other benchmarks

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+                             bibtex=BIBTEX)
+
+        self._stimulus_set = stimulus_set
+        self._target_assembly = average_repetition(assembly)
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
+        self._metric = metric
+
+        self.bootstrap_samples = bootstrap_samples
+        self.num_sample_arrs = num_sample_arrays_candidate
+        self._arr_size_mm = (np.ptp(self._target_assembly.neuroid.x.data),
```
mschrimpf, line 50: you might just want to write out `array`, you don't save that many characters by removing `ay` ;-)

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+                             parent='IT',
+                             bibtex=BIBTEX)
+
+        self._stimulus_set = stimulus_set
+        self._target_assembly = average_repetition(assembly)
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
+        self._metric = metric
+
+        self.bootstrap_samples = bootstrap_samples
+        self.num_sample_arrs = num_sample_arrays_candidate
```
mschrimpf, line 49: comment, e.g. "How many virtual Utah array placement locations to sample"

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+        self._arr_size_mm = (np.ptp(self._target_assembly.neuroid.x.data),
+                             np.ptp(self._target_assembly.neuroid.y.data))
+
+    def __call__(self, candidate):
+
+        candidate.start_recording(recording_target='IT', time_bins=[(70,170)])
```
mschrimpf, line 55: missing space after 70. PyCharm has an autoformat feature which could be handy

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+
+        pairwise_tissue_corr_candidate = []
+        for window in sample_array_locations(activations.neuroid, self.num_sample_arrs, self._arr_size_mm):
+            pairwise_tissue_corr_candidate.append(sample_response_corr_vs_dist(activations[window],
+                                                              int(self.bootstrap_samples/self.num_sample_arrs)))
+
+        return self.score(np.hstack(pairwise_tissue_corr_candidate))
+
+    def create_tissue_response_correlation_target(self):
+
```
mschrimpf, line 67: remove newline

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+        pairwise_tissue_corr_target = []
+        for animal in set(self._target_assembly.neuroid.animal.data):
+            for arr in set(self._target_assembly.neuroid.arr.data):
+
+                sub_assembly = self._target_assembly.sel(animal=animal, arr=arr)
+                bootstrap_samples_sub_assembly = int(self.bootstrap_samples*(sub_assembly.neuroid.size/
```
mschrimpf, line 73: missing space before and after `*`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+                                                              int(self.bootstrap_samples/self.num_sample_arrs)))
+
+        return self.score(np.hstack(pairwise_tissue_corr_candidate))
+
+    def create_tissue_response_correlation_target(self):
+
+        pairwise_tissue_corr_target = []
+        for animal in set(self._target_assembly.neuroid.animal.data):
+            for arr in set(self._target_assembly.neuroid.arr.data):
+
```
mschrimpf, line 71: remove newline

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+        activations = make_static(activations)
+
+        pairwise_tissue_corr_candidate = []
+        for window in sample_array_locations(activations.neuroid, self.num_sample_arrs, self._arr_size_mm):
+            pairwise_tissue_corr_candidate.append(sample_response_corr_vs_dist(activations[window],
+                                                              int(self.bootstrap_samples/self.num_sample_arrs)))
+
+        return self.score(np.hstack(pairwise_tissue_corr_candidate))
+
+    def create_tissue_response_correlation_target(self):
```
mschrimpf, line 66: in general: add some more comments to what your functions do

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+    def score(self, candidate_statistic):
+
+        self._target_statistic = self._target_statistic_fn()
+
+        def bin_indices():
+            t = np.where(np.logical_and(self._target_statistic[0] >= lower_bound_mm,
+                                        self._target_statistic[0] < lower_bound_mm+0.1))[0]
+            c = np.where(np.logical_and(candidate_statistic[0] >= lower_bound_mm,
+                                        candidate_statistic[0] < lower_bound_mm + 0.1))[0]
+            return t, c, t.size>30 & c.size>30 # TODO 30 neurons per bin enough for stable estimate?
```
mschrimpf, line 91: todo comment

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+
+        bin_scores = []
+        arr_diag_mm = round(np.linalg.norm(self._arr_size_mm), 1)
+        for lower_bound_mm in np.linspace(0, arr_diag_mm, int(arr_diag_mm*10+1)):
+            in_bin_t, in_bin_c, enough_data = bin_indices()
+            if enough_data: # TODO only evaluate on bins with min number of neurons for stability?
```
mschrimpf, line 97: todo comment

brainscore/benchmarks/__init__.py
```diff
@@ -193,6 +193,8 @@ def _experimental_benchmark_pool():
     pool['dicarlo.SanghaviMurty2020.IT-pls'] = LazyLoad(DicarloSanghaviMurty2020ITPLS)
     from .rajalingham2020 import DicarloRajalingham2020ITPLS
     pool['dicarlo.Rajalingham2020.IT-pls'] = LazyLoad(DicarloRajalingham2020ITPLS)
+    from .Lee2020 import DicarloLee2020ITTissueResponseCorrelation
+    pool['dicarlo.Lee2020.IT-tissue'] = LazyLoad(DicarloLee2020ITTissueResponseCorrelation)
```
mschrimpf, line 197: I would call it `response_correlation` instead of `tissue`.
Likewise, I would rename the benchmark to `DicarloLee2020ITResponseCorrelation`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,181 @@
...
+                bootstrap_samples_sub_assembly = int(self.bootstrap_samples*(sub_assembly.neuroid.size/
+                                                                             self._target_assembly.neuroid.size))
+
+                pairwise_tissue_corr_target.append(sample_response_corr_vs_dist(sub_assembly,
+                                                                                bootstrap_samples_sub_assembly,
+                                                                                self.ceiling))
+
+        return np.hstack(pairwise_tissue_corr_target)
+
+    def score(self, candidate_statistic):
```
mschrimpf, line 82: this entire method (and methods it uses) should actually be a **metric** so that multiple benchmarks can re-use it on different data.

Maybe it would also make sense to move the characterization (or property) functions into a common directory so that they too can be re-used.  But that doesn't have to happen here, it can be a thing in case another benchmark wants to re-use your code.

mschrimpf (commented):
Being nitpicky especially because this is your first PR and because I know you're looking for software-engineering feedback, but overall this looks great!
I think you should run PyCharm's autoformat on all your code, that will resolve a lot of my comments already.

brainscore/benchmarks/__init__.py
mschrimpf: yea I think this gets way too long. Maybe we can talk about it in our meeting, ideally we could find a name that is:
* short
* mentions something spatial (about the tissue)
* mentions something about the response correlation
But not sure if we can find an identifier that satisfies all three

pmcgrath249:
Thank you this was great input, I appreciate nitpickiness :)

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,192 @@
+import brainscore
+import numpy as np
+from brainscore.benchmarks import BenchmarkBase
+from brainscore.benchmarks._neural_common import average_repetition
+from brainscore.benchmarks.perturbation_prototypes import Rajalingham2019
```
mschrimpf, line 5: use MajajHong2015

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,192 @@
...
+        '''
+        This benchmark compares the distribution of pairwise response correlation as a function of distance between the
+            MajajHong assembly and a candidate assembly
+        Plots of this can be found in Figure 2 of the corresponding publication (see BIBTEX)
+        '''
+        stimulus_assembly = Rajalingham2019()._target_assembly
+        stimulus_set = stimulus_assembly.stimulus_set
+        assembly = brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT')
+        assembly = make_static(assembly)
+        assembly = assembly[:, np.in1d(assembly.presentation.image_id, list(set(stimulus_set['image_id'])))]
```
mschrimpf, line 36: use MajajHong2015

let's keep the exploratory benchmark for Rajalingham2019 stimuli e.g. in direct-causality

brainscore/benchmarks/__init__.py
pmcgrath249: rename identifier to spatial_correlation and SpatialCorrelationSimilarity

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,188 @@
...
+        self._score = SpatialCorrelationSimilarity(ks_similarity, 1)
+
+        self.bootstrap_samples = 100_000
+        self.num_sample_arrs = 10  # number of simulated Utah arrays sampled from candidate model tissue
+        self._array_size_mm = (np.ptp(self._target_assembly.neuroid.x.data),  # physical size of Utah array in mm
+                               np.ptp(self._target_assembly.neuroid.y.data))
+
+    def __call__(self, candidate):
+        '''
+        :param candidate: NeuroidAssembly
```
mschrimpf, line 52: candidate should be of type `BrainModel`

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,188 @@
...
+        :return: Score for the pairwise response correlation compared to the MajajHong Assembly
+        '''
+        candidate.start_recording(recording_target='IT', time_bins=[(70, 170)])
+        activations = candidate.look_at(self._stimulus_set)
+        activations = make_static(activations)
+
+        pairwise_tissue_corr_candidate = []
+        for window in sample_array_locations(activations.neuroid, self.num_sample_arrs, self._array_size_mm):
+            pairwise_tissue_corr_candidate.append(sample_response_corr_vs_dist(activations[window],
```
mschrimpf, line 61: formatting went a bit overboard here, I guess a newline after `append` (and then auto-format again)

mschrimpf: you might also want to separate this into two lines to make it easier to read, i.e. put the `sample_response_corr_vs_dist` result in a variable that you pass to the list

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,188 @@
...
+        '''
+        This benchmark compares the distribution of pairwise response correlation as a function of distance between the
+            MajajHong assembly and a candidate assembly
+        Plots of this can be found in Figure 2 of the corresponding publication (see BIBTEX)
+        '''
+        assembly = brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT')
+        assembly = make_static(assembly)
+
+        super(self).__init__(identifier='dicarlo.Lee2020.IT-tissue',
+                             ceiling_func=lambda: InternalConsistency()(assembly.transpose('presentation', 'neuroid')),
```
mschrimpf, line 35: we'll probably want a different ceiling here that reflects the differences in spatial correlations

mschrimpf: Effectively there are then two ceilings like we talked about: one for the internal consistency to normalize correlation values, and one for e.g. cross-animal consistency to normalize model scores

brainscore/benchmarks/Lee2020.py
```diff
@@ -0,0 +1,188 @@
...
+def pairwise_distances(assembly):
+    '''
+    Convenience function creating a simple lookup table for pairwise distances
+    :param assembly: NeuroidAssembly
+    :return: square matrix where each entry is the distance between the neuroids at the corresponding indices
+    '''
+    if hasattr(assembly, 'tissue_x'):
+        locations = np.stack([assembly.neuroid.tissue_x.data, assembly.neuroid.tissue_y.data]).T
+    elif hasattr(assembly, 'x'):
+        locations = np.stack([assembly.neuroid.x.data, assembly.neuroid.y.data]).T
```
mschrimpf, line 134: let's keep it at only one way to do things, `tissue_x`

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,54 @@
+import numpy as np
+from brainscore.metrics import Score, Metric
+
+
+def _score(scores):
```
mschrimpf, line 5: maybe rename to aggregate?

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def __init__(self, metric, bin_size_mm):
+        '''
+        Computes the similarity of two given distributions using a given metric. The metric is applied to each bin which
+        in turn are created based on a given bin size and the independent variable of the distributions
+
+        :param metric: metric to be applied to each bin
+        :param bin_size_mm: size per bin in mm | one fixed size, utilize Score.RAW_VALUES_KEY to change weighting
+        '''
+        self._metric = metric
```
mschrimpf, line 24: maybe this should be called `similarity_function` to not overload the metric term

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,54 @@
...
+        self._bin_size = bin_size_mm
+
+    def __call__(self, target_statistic, candidate_statistic):
+        '''
+        :param target_statistic: list of 2 lists, [0] holds the independent variable -> binning over this
+        :param candidate_statistic: list of 2 lists, [0] holds the independent variable -> binning over this
```
mschrimpf, line 30: needs more explanation of what is in the lists

tests/test_benchmarks/test_lee2020.py
```diff
@@ -0,0 +1,46 @@
...
+
+    if test == 'extreme':  # extreme r=0, std=1
+        assert np.abs(np.mean(r)) < 0.01
+        assert np.std(r) - 1 < 0.01
+    elif test == 'rand':  # rand r=0 w/ std small
+        assert np.abs(np.mean(r)) < 0.01
+        assert np.std(r) < .2
+    elif test == 'max':  # same r=1 w/ std=0
+        assert np.abs(np.mean(r)) - 1 < 0.01
+        assert np.std(r) < 0.01
```
mschrimpf, line 33: just separate into separate functions

brainscore/benchmarks/Lee2020.py
pmcgrath249: I am using this for both, the assembly and the model. Do you want me to add `tissue_x` to the assembly before computing? Or would you prefer to separate functions?

brainscore/metrics/spatial_correlation.py
pmcgrath249: How about: `list of 2 lists, [0] distances -> binning over this, [1] correlation per distance value`

pmcgrath249:
This last commit is the first attempt at a CrossIndividualValidationCeiling. I was not sure how inheritance in this case works because I could make this work with your Transformation class, at least I think, but after giving it some thought I found it a little counter-intuitive as it should be a Ceiling.

brainscore/metrics/IndividualCeiling.py
```diff
@@ -0,0 +1,56 @@
...
+    def __init__(self, benchmark, modify_benchmark=lambda benchmark, target: None):
+        self._benchmark = benchmark
+        if modify_benchmark:
+            self._modify_benchmark = modify_benchmark
+        self._set_generator = InterIndividualCrossValidation()
+
+    def __call__(self, assemblies):
+        scores = []
+        for target, source in self._set_generator(assemblies):
+            self._modify_benchmark(self._benchmark, target)  # TODO too hacky??
```
pmcgrath249, line 28: Tried to make it generic but seems a little hacky, right?

brainscore/metrics/IndividualCeiling.py
```diff
@@ -0,0 +1,56 @@
+from brainscore.metrics.ceiling import Ceiling
+from brainscore.metrics.transformations import Transformation
+from brainscore.benchmarks._neural_common import average_repetition
+import numpy as np
+import xarray as xr
+
+
+class InterIndividualCrossValidation:  # TODO inheritance Transformation???: InterIndivCrossVal = pipe, Ceiling = apply?
```
pmcgrath249, line 8: I was not sure how inheritance in this case works because I could make this work with your `Transformation `class, at least I think, but after giving it some thought I found it a little counter-intuitive as `InterIndividualBenchmarkCeiling` should be a `Ceiling`.

brainscore/metrics/IndividualCeiling.py
```diff
@@ -0,0 +1,56 @@
+from brainscore.metrics.ceiling import Ceiling
+from brainscore.metrics.transformations import Transformation
+from brainscore.benchmarks._neural_common import average_repetition
+import numpy as np
+import xarray as xr
+
+
+class InterIndividualCrossValidation:  # TODO inheritance Transformation???: InterIndivCrossVal = pipe, Ceiling = apply?
+
+    def __call__(self, assembly):  # TODO add support for list of assemblies
```
pmcgrath249, line 10: Would adding support for multiple assemblies make sense? Would most benchmarks be able to cope with assemblies having different sizes in the `presentation ` dimension and being concatenated along the `neuroid `dimension?

brainscore/metrics/IndividualCeiling.py
```diff
@@ -0,0 +1,56 @@
...
+        for target, source in self._set_generator(assemblies):
+            self._modify_benchmark(self._benchmark, target)  # TODO too hacky??
+
+            source = average_repetition(source)
+            scores.append(self._benchmark(source))
+
+        return np.mean([s[0].data for s in scores]), scores
+
+
+def main():
```
pmcgrath249, line 36: please ignore all this and the fact that it is not in ceiling.py

brainscore/metrics/IndividualCeiling.py
mschrimpf: I think you could directly use the `CrossValidation` class inside of the ceiling class. If you set the type to kfold and the number of splits to equal to the number of subjects, it should always split the assembly into {pool, heldout} per split. https://github.com/mschrimpf/brain-score/blob/ebd6af2ce73e9837665a9b874e423ff875470683/brainscore/metrics/behavior_differences.py#L92-L93 is similar to that, holding out one task per split (instead of subject)

brainscore/metrics/IndividualCeiling.py
mschrimpf: Yes, `presentation x neuroid` is the default assembly format. But don't worry too much about making it work for everything, let's first get it to run for this particular benchmark and then we can still think about generifying.

brainscore/metrics/IndividualCeiling.py
mschrimpf: yes I also don't quite understand why you need to modify the benchmark -- wouldn't you be able to apply the metric? I.e. `score = metric(source, target)`.

pmcgrath249: true, I did not think of that...

brainscore/metrics/IndividualCeiling.py
pmcgrath249: I think the problem is that the statistics are not assemblies anymore but np.arrays. Do you think it is worth converting?

brainscore/benchmarks/Lee2020.py
mschrimpf: ah -- yes let's set `assembly['tissue_x'] = assembly['x']` before computing and @jjpr-mit and I will try to update the assemblies accordingly

brainscore/metrics/IndividualCeiling.py
mschrimpf: Up to you -- I have kept things in assembly format just because I find it easier to work with labeled data, but you're building the benchmark, so you can decide what level of metadata you're comfortable with

brainscore/benchmarks/__init__.py
```diff
@@ -193,6 +193,8 @@ def _experimental_benchmark_pool():
     pool['dicarlo.SanghaviMurty2020.IT-pls'] = LazyLoad(DicarloSanghaviMurty2020ITPLS)
     from .rajalingham2020 import DicarloRajalingham2020ITPLS
     pool['dicarlo.Rajalingham2020.IT-pls'] = LazyLoad(DicarloRajalingham2020ITPLS)
+    from .lee2020 import DicarloLee2020ITSpatialCorrelation
+    pool['dicarlo.Lee2020.IT-spatial_correlation'] = LazyLoad(DicarloLee2020ITSpatialCorrelation)
```
mschrimpf, line 197: I think this should be `dicarlo.MajajHong2015.IT-spatial_correlation` because that is where the data came from. Ideally the characterization in Lee et al. 2020 as well as you as the builder of comparison metric should be featured somehow as well; for now that should go into the Metric description I think.

This is a discussion we should also have with Jim. My preference is to credit: 1) data provider (from the paper that the data was uploaded to brain-score from), 2) metric builder (paper where the metric for brain-model comparison was introduced), 3) benchmark implementer

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+            URL = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116},
+            eprint = {https://www.biorxiv.org/content/early/2020/07/10/2020.07.09.185116.full.pdf},
+            journal = {bioRxiv}}'''
+
```
mschrimpf, line 21: as above, I think this should point to the data (import the MajajHong BIBTEX)

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+
+class DicarloLee2020ITSpatialCorrelation(BenchmarkBase):
```
mschrimpf, line 23: should accordingly also change name

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+
+    def __init__(self):
+        '''
+        This benchmark compares the distribution of pairwise response correlation as a function of distance between the
+            MajajHong assembly and a candidate assembly
```
mschrimpf, line 28: MajajHong --> MajajHong2015
candidate assembly --> candidate BrainModel

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+        Plots of this can be found in Figure 2 of the corresponding publication (see BIBTEX)
+        '''
+        super().__init__(identifier='dicarlo.Lee2020.IT-spatial_correlation',
```
mschrimpf, line 31: change identifier accordingly

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+    # create index masks of neuroids within sample windows
+    for i in range(num_sample_arrs):
+        yield np.logical_and.reduce([neuroid.tissue_x.data <= upper_corner[i, 0],
+                                     neuroid.tissue_x.data >= lower_corner[i, 0],
+                                     neuroid.tissue_y.data <= upper_corner[i, 1],
+                                     neuroid.tissue_y.data >= lower_corner[i, 1]])
+
+
+def tissue_update(assembly):
+    # temporary function until assemblies updated
```
mschrimpf, line 186: what does the function do? What specific update needs to happen for this method to become obsolete?

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+class DicarloLee2020ITSpatialCorrelation(BenchmarkBase):
+
+    def __init__(self):
+        '''
+        This benchmark compares the distribution of pairwise response correlation as a function of distance between the
+            MajajHong assembly and a candidate assembly
+        Plots of this can be found in Figure 2 of the corresponding publication (see BIBTEX)
+        '''
+        super().__init__(identifier='dicarlo.Lee2020.IT-spatial_correlation',
+                         ceiling_func=None,
```
mschrimpf, line 32: ceiling missing

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+                         bibtex=BIBTEX)
+
+        assembly = brainscore.get_assembly('dicarlo.MajajHong2015').sel(region='IT')
+        assembly = make_static(assembly)
+        assembly = tissue_update(assembly)
+
+        self._neuroid_reliability = InternalConsistency()(assembly.transpose('presentation', 'neuroid'))
+        self._stimulus_set = assembly.stimulus_set
+        self._target_assembly = average_repetition(assembly)
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
```
mschrimpf, line 44: why this rename of the function, could just replace all uses of `self._target_statistic_fn` with `self.create_tissue_response_correlation_target` no?

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+
+        self.bootstrap_samples = 100_000
+        self.num_sample_arrs = 10  # number of simulated Utah arrays sampled from candidate model tissue
+        self._array_size_mm = (np.ptp(self._target_assembly.neuroid.tissue_x.data),  # physical size of Utah array in mm
+                               np.ptp(self._target_assembly.neuroid.tissue_y.data))
+
+    def __call__(self, candidate_model):
+        '''
+        This computes the statistics, i.e. the pairwise response correlation of candidate and target, respectively and
+        compute a score based on the ks similarity of the two resulting distributions
```
mschrimpf, line 55: compute -> computes

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+        self._target_assembly = average_repetition(assembly)
+        self._target_statistic_fn = self.create_tissue_response_correlation_target
+        self._score = SpatialCorrelationSimilarity(similarity_function=inv_ks_similarity, bin_size_mm=.1)
+
+        self.bootstrap_samples = 100_000
+        self.num_sample_arrs = 10  # number of simulated Utah arrays sampled from candidate model tissue
+        self._array_size_mm = (np.ptp(self._target_assembly.neuroid.tissue_x.data),  # physical size of Utah array in mm
+                               np.ptp(self._target_assembly.neuroid.tissue_y.data))
+
+    def __call__(self, candidate_model):
```
mschrimpf, line 52: suggest shortening to `candidate` to remove the focus on "model" and adding a type hint
`candidate_model` -> `candidate: BrainModel`

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+
+        candidate_statistic = []
+        for window in sample_array_locations(candidate_assembly.neuroid, self.num_sample_arrs, self._array_size_mm):
+            array_statistic = sample_response_corr_vs_dist(candidate_assembly[window],
+                                                           int(self.bootstrap_samples / self.num_sample_arrs))
+            candidate_statistic.append(array_statistic)
+
+        self._target_statistic = self._target_statistic_fn()
+        self._ceiling_func = lambda: InterIndividualStatisticsCeiling(self._score)(
+            self._target_statistic)  # TODO integrate ceiling
```
mschrimpf, line 71: todo

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+        '''
+        candidate_model.start_recording(recording_target='IT', time_bins=[(70, 170)])
+        candidate_assembly = candidate_model.look_at(self._stimulus_set)
+        candidate_assembly = make_static(candidate_assembly)
+
+        candidate_statistic = []
+        for window in sample_array_locations(candidate_assembly.neuroid, self.num_sample_arrs, self._array_size_mm):
+            array_statistic = sample_response_corr_vs_dist(candidate_assembly[window],
+                                                           int(self.bootstrap_samples / self.num_sample_arrs))
+            candidate_statistic.append(array_statistic)
```
mschrimpf, line 67: this (together with `np.hstack(candidate_statistic)`) should go inside its own function, e.g. `sample_local_tissue_statistics` or so

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+
+        return target_statistic
+
+
+def sample_response_corr_vs_dist(assembly, num_samples, neuroid_reliability=None):
+    '''
+    1. Samples random pairs from the assembly
+    2. Computes distances for all pairs
+    3. Computes the response correlation between items of each pair
+    (4. Ceils the response correlaions)
```
mschrimpf, line 102: correlaions --> correlations

add: if `neuroid_reliability` is not None

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+    :param assembly: NeuroidAssembly without stimulus repetitions
+    :param num_samples: how many random pair you want to be sampled out of the data
+    :param neuroid_reliability: if not None: expecting Score object containing reliability estimates of all neuroids
+    :return: [distance, pairwise_correlation_of_neuroids], pairwise correlations can be ceiled
+    '''
+    rng = np.random.default_rng()
```
mschrimpf, line 108: I would use a random state here (and whenever you use randomness) to ensure perfect reproducibility:
`rng = RandomState(0)`. Need to `from numpy.random import RandomState`, and 0 can be any number to seed the generator.

brainscore/benchmarks/lee2020.py
```diff
@@ -0,0 +1,208 @@
...
+def sample_array_locations(neuroid, num_sample_arrs, array_size_mm):
+    '''
+    Generator: Sample Utah array-like portions from artificial model tissue and generate masks
+    :param neuroid: NeuroidAssembly.neuroid
+    :param num_sample_arrs: number of simulated Utah arrays you want to sample
+    :param array_size_mm: Tuple, physical size of the arrays you want to sample in mm
+    :return: list of masks in neuroid dimension of assembly, usage: assembly[mask] -> neuroids within one array
+    '''
+    bound_max_x, bound_max_y = np.max([neuroid.tissue_x.data, neuroid.tissue_y.data], axis=1) - array_size_mm
+    rng = np.random.default_rng()
```
mschrimpf, line 170: use RandomState(0)

brainscore/metrics/inter_individual_stats_ceiling.py
```diff
@@ -0,0 +1,33 @@
...
+def _aggregate(scores):
+    center = np.mean(scores)
+    error = np.std(scores)
+    aggregate_score = Score([center, error], coords={'aggregation': ['center', 'error']}, dims=('aggregation',))
+    aggregate_score.attrs[Score.RAW_VALUES_KEY] = scores
+
+    return aggregate_score
+
+
+class InterIndividualStatisticsCeiling(Ceiling):
```
mschrimpf, line 14: would add some comments here of what this does

brainscore/metrics/inter_individual_stats_ceiling.py
```diff
@@ -0,0 +1,33 @@
...
+    def __init__(self, metric):
+        '''
+        Cross-validation-like, animal-wise computation of ceiling
+        :param metric: used to compute the ceiling
+        '''
+        self.metric = metric
+
+    def __call__(self, target_statistic):
+        '''
+
```
mschrimpf, line 24: unnecessary newline

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,55 @@
...
+def _aggregate(scores):
+    center = np.mean(scores)
+    error = np.std(scores)
+    aggregate_score = Score([center, error], coords={'aggregation': ['center', 'error']}, dims=('aggregation',))
+    aggregate_score.attrs[Score.RAW_VALUES_KEY] = scores
+
+    return aggregate_score
+
+
+class SpatialCorrelationSimilarity(Metric):
```
mschrimpf, line 14: describe what this does (could e.g. pull up the comments from `__init__`)

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,55 @@
...
+        :param candidate_statistic: list of 2 lists, [0] distances -> binning over this, [1] correlation per distance value
+        '''
+        self.target_statistic = target_statistic
+        self.candidate_statistic = candidate_statistic
+        self._bin_min = np.min(self.target_statistic[0])
+        self._bin_max = np.max(self.target_statistic[0])
+
+        bin_scores = []
+        for in_bin_t, in_bin_c, enough_data in self._bin_masks():
+            if enough_data:  # TODO only evaluate on bins with min number of neurons for stability?
```
mschrimpf, line 40: todo

brainscore/metrics/spatial_correlation.py
```diff
@@ -0,0 +1,55 @@
...
+                bin_scores.append(self.similarity_function(self.target_statistic[1, in_bin_t],
+                                                           self.candidate_statistic[1, in_bin_c]))
+
+        return _aggregate(bin_scores)
+
+    def _bin_masks(self):
```
mschrimpf, line 46: what does this do?

tests/test_benchmarks/test_lee2020.py
```diff
@@ -0,0 +1,49 @@
+import pytest
+import brainscore
+from brainscore.benchmarks import lee2020
+import numpy as np
+
+
+def rand_corr_data():
+    return np.random.rand(5000 * 168).reshape((168, 5000))
```
mschrimpf, line 8: use RandomState(0)

mschrimpf (commented):
Aside from the stylistic comments, the only major comment I have is that there should be a benchmark test based on actual activations. See e.g. https://github.com/brain-score/brain-score/blob/0e34c8389d912ae7be9c4012abdb52e4a62c51ce/tests/test_benchmarks/test___init__.py#L160, you can probably even use the same precomputed features

brainscore/benchmarks/lee2020.py
pmcgrath249: if I am [understanding correctly for numpy > 1.17](https://numpy.org/doc/stable/reference/random/generator.html#numpy.random.Generator) the preferred way is to use `np.random.default_rng()` to which you can pass a seed making everything reproducible. This should result in the desired behavior, right?

Question for this: In case you like that solution, do you think it makes sense to have the random generator being part of each benchmark instance? Would be more convenient but also if people play around and call things in different orders, random numbers are not going to be  the same.

brainscore/benchmarks/lee2020.py
pmcgrath249: oh good, this was a reminder. as you noticed, no ceiling functions is passed in `__init__`. my reasoning here was that we are computing the ceiling based on the target statistic which would only be computed in ` __call__`. Do you want me to change that or is it ok to only compute the ceiling at this time in the computations?
In case ceiling is stored anyways, I should probably move it into `__init__`

brainscore/benchmarks/__init__.py
pmcgrath249: do you want me to move the whole benchmark code to brainscore/benchmarks/majajhong2015 then?

brainscore/benchmarks/lee2020.py
mschrimpf: oh I didn't know about the new Generators, neat. 
I think where you have the `rng` now is fine, all that is relevant is that if you call the benchmark with the same parameters (aka model), it should yield the exact identical result.

brainscore/benchmarks/lee2020.py
pmcgrath249: For completeness: We are going to use LazyLoad here which allows us to put the function into `__init__`

mschrimpf:
So in terms of code organization what @jjpr and I think makes most sense is the following:
1. Group benchmarks by assembly name (e.g. `majajhong2015.py`) as we do now. Define all benchmarks on that data within this file
2. Put each benchmark in a class within the file (e.g. `class DicarloMajajHong2015ITSpatialCorrelation`) and put all functions that are not shared inside this class (there might be issues with this down the line if a benchmark ever mixes multiple datasets, but let's follow YAGNI)

pmcgrath249:
I do love me some YAGNI! Thanks!

pmcgrath249:
going to create a new PR (referencing this thread)

PR #278, opened 2021-06-25
Use BrainIO

jjpr-mit:
Use BrainIO as a library with a local `lookup.csv`.

- [ ] before merging: pull the latest `lookup.csv` to make sure we're not leaving any PRs behind


test_setup.sh
```diff
@@ -8,5 +8,5 @@ do
   aws --no-sign-request s3 cp s3://brain-score-tests/tests/test_benchmarks/${f} ${SCRIPT_DIR}/tests/test_benchmarks/
 done
 
-pip install git+https://github.com/brain-score/model-tools --no-deps
+pip install git+https://github.com/brain-score/model-tools@core_refactor --no-deps
```
mschrimpf, line 11: once this qualification is removed, I think this looks good

jjpr-mit: Of course, it’ll fail tests after that.

mschrimpf: so in what order do you want to merge? 1. brain-score, 2. model-tools?

jjpr: That works. Then I'll do a separate PR to remove the qualification.

mschrimpf (approved):
Approving assuming that you will complete the TODO in the PR: `pull the latest lookup.csv to make sure we're not leaving any PRs behind` (e.g. https://github.com/brain-score/brainio_collection/pull/46 was just merged)

PR #280, opened 2021-07-15
Migrated from https://github.com/brain-score/brainio_collection.

jjpr-mit:


mschrimpf:
I think the `scripts/` folder is unnecessary. Let's move everything within that directory up by one level.
Ideally we would also move all the notebooks into their corresponding directories, but it's fine if we don't do that within this PR.

jjpr-mit:
I think it keeps things tidier.  And it's only going to grow over time.

mschrimpf:
let me rephrase: I think `brain-score/packaging/scripts/` is unnecessary, I find `brain-score/packaging/` completely sufficient for what is in there.

PR #305, opened 2021-10-15
Catalog path

jjpr-mit:
Add path for lookup.csv to lookup DataFrame so packaging can write to it.


mschrimpf:
re-trigger after merging https://github.com/brain-score/brainio/pull/11

PR #310, opened 2021-12-08
Added support for 2022 competition

mike-ferguson:
This PR adds the fields into evaluation.py and competition.py in order to enable competition database support.

### DIFF
--- a/brainscore/submission/configuration.py
+++ b/brainscore/submission/configuration.py
@@ -40,10 +40,11 @@ class SubmissionConfig(BaseConfig):
     Configuration properties for newly submitted models, which also have a submission entry in database.
     """
 
-    def __init__(self, model_type, user_id, jenkins_id, public, **kwargs):
+    def __init__(self, model_type, user_id, jenkins_id, public: bool, competition_submission: str, **kwargs):
         super(SubmissionConfig, self).__init__(jenkins_id=jenkins_id, **kwargs)
         self.submission = Submission.create(id=jenkins_id, submitter=user_id, timestamp=datetime.datetime.now(),
                                             model_type=model_type, status='running')
+        self.competition_submission = competition_submission
         self.public = public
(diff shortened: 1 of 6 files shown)


brainscore/submission/configuration.py
```diff
@@ -60,4 +61,6 @@ def object_decoder(config, work_dir, config_path, db_secret, jenkins_id):
         return SubmissionConfig(model_type=config['model_type'], user_id=config['user_id'], work_dir=work_dir,
                                 config_path=config_path,
                                 jenkins_id=jenkins_id, db_secret=db_secret,
-                                public=config['public'] == "True")
+                                public=config['public'] == "True",
+                                competition_submission=config['competition'] == "cosyne2022")
```
mschrimpf, line 65: I thought we were going to write the competition name, i.e. `"cosyne2022"` and not a bool?

brainscore/submission/evaluation.py
```diff
@@ -70,7 +70,8 @@ def run_evaluation(config_dir, work_dir, jenkins_id, db_secret, models=None,
                 for model_name in test_models:
                     model_entry, created = Model.get_or_create(name=model_name, owner=submission_entry.submitter,
                                                                defaults={'public': submission_config.public,
-                                                                         'submission': submission_entry})
+                                                                         'submission': submission_entry,
+                                                                         'competition': submission_config.competition_submission})
```
mschrimpf, line 74: if I understand the current flow right, this will write a bool rather than `"cosyne2022"`. I think we should write the name of the competition though

mschrimpf (changes requested):

mschrimpf (approved):
looks good to me, if tests pass

mschrimpf:
re-trigger, I think what happened was that this PR + #311 ran at the same time

PR #314, opened 2022-01-07
Added competition CharField to submission fields; added two tests; changed a fixture

mike-ferguson:
This PR adds support for the competition, allowing for the database to be written to with the correct competition field. Also adds two tests to confirm the competition field is set (both when it is cosyne2022 and None). Changes a fixture's field.


tests/test_submission/test_integration.py
```diff
@@ -44,6 +44,29 @@ def teardown_method(self):
...
+        config_dir = str(os.path.join(os.path.dirname(__file__), 'configs/'))
+        run_evaluation(config_dir, working_dir, 33, TestIntegration.database, models=['alexnet'],
+                       benchmarks=['dicarlo.MajajHong2015.IT-pls'])
+        model = Model.get()
+        assert model.competition == "cosyne2022"
+
+    def test_competition_field_none(self, tmpdir):
+        working_dir = str(tmpdir.mkdir('sub'))
+        config_dir = str(os.path.join(os.path.dirname(__file__), 'configs/'))
+        submission = Submission.create(id=33, submitter=1, timestamp=datetime.now(),
```
mschrimpf, line 58: isn't this submission changed to have the `competition` field set? https://github.com/brain-score/brain-score/pull/314/files#diff-5bb7cee62807d0b6b68c098d0ca440824cac9f167b7681d48f3055d009da48a8R91

tests/test_submission/test_integration.py
```diff
@@ -44,6 +44,29 @@ def teardown_method(self):
...
+                                       model_type='BaseModel', status='running')
+        model = Model.create(name='alexnet', owner=submission.submitter, public=False,
+                             submission=submission)
+        with open(f'{config_dir}submission_40.json', 'w') as rerun:
```
mschrimpf, line 62: probably not essential for this test, but it's confusing that at first the id is 33 and now 40

tests/test_submission/test_integration.py
```diff
@@ -44,6 +44,29 @@ def teardown_method(self):
...
+            rerun.write(f"""{{
+                "model_ids": [{model.id}], "user_id": 1, "competition": null}}""")
+        run_evaluation(config_dir, working_dir, 40, TestIntegration.database,
```
mschrimpf, line 65: as above, now using id 40 instead of 33. Maybe create another submission zip submission_40.zip, copied from submission_33?

mschrimpf (approved):

PR #328, opened 2022-06-03
rename image_id -> stimulus_id in I2n.target_distractor_scores

mschrimpf:


mschrimpf:
re-trigger tests after merging https://github.com/brain-score/brainio/pull/19

mschrimpf:
There is circularity in the merging order now:
* this PR depends on https://github.com/brain-score/model-tools/pull/62 because `tests.test_submission.test_integration.TestIntegration.test_rerun_evaluation` and `tests.test_submission.test_integration.TestIntegration.test_evaluation` fail due to model-tools using `get_image` (which is fixed in the model-tools PR)
* https://github.com/brain-score/model-tools/pull/62 fails because of fixes that should exist here
I'm thus going to merge this PR, and then re-trigger the model-tools PR.

PR #329, opened 2022-06-08
Bug in TestPrecomputed

jjpr-mit:
When I updated test_benchmarks.test___init__.TestPrecomputed.run_test to use from_files, it broke.


mschrimpf:
redundant with #328

PR #330, opened 2022-06-17
let ErrorConsistency Score output have center and error

mschrimpf:
Otherwise the submission evaluation fails because it wants to write `score.sel(aggregation='center')` to the database, see e.g. http://braintree.mit.edu:8080/job/run_benchmarks/2484/console

Choose std over condition and subject for error. The original paper did not have error estimates as far as I can tell


mike-ferguson (approved):

mschrimpf:
re-trigger due to weird segfault in unit test..
image

PR #344, opened 2022-10-18
fix PI name for Hermann2020 benchmarks

mschrimpf:
Brendel is not on the paper, last author is Kornblith (https://papers.nips.cc/paper/2020/hash/db5f9f42a7157abe65bb145000b5871a-Abstract.html)
image


mike-ferguson (approved):

mschrimpf:
openmind disk quota exceeded during jenkins test, re-triggering

PR #356, opened 2023-01-27
Jacob 2020 Behavioral Benchmarks: 3D Processing

mike-ferguson:
This PR will contain Jacob 2020 3D Processing Index benchmark:

Fundamentally, humans and models are performing different tasks in this benchmark. Models are passively viewing
a stimuli (and having those activations retrieved from IT), whereas humans performed a visual search task
in the papers listed below. The reasons for doing so are to replicate benchmark work done in Jacob et al 2020.

Following Jacob et al. 2020, we retrieve internal neural activity from the model (layer IT)
rather than having the model perform the same visual search task that humans were performing
in the behavioral experiment from which we take the behavioral data to compare against (below):


mschrimpf:
if the PR isn't ready yet, can you please mark it as a draft PR?

PR #359, opened 2023-03-09
use new jenkins_id instead of id for submissions

mike-ferguson:
Allows submission system to now write the newly updated `jenkins_id` column, as well as the auto-incrementing `id`. See https://github.com/brain-score/brain-score.web/pull/159 for more info


mschrimpf (changes requested):
The [database models specification](https://github.com/brain-score/brain-score/blob/1b8062558da7f904d9a10c718f6b534dd94a0f0d/brainscore/submission/models.py#L72-L73) also needs to be addressed. I'm hoping tests will catch this as well and fail.

mschrimpf (approved):

mschrimpf:
Some issue with Jenkins tests, but submissions are failing, so I'll merge this

PR #360, opened 2023-03-20
Islam2021

GaspTO:


GaspTO:
I think Travis is failing because it can't download the stimulus set due to lack of permissions? 
I don't think this is a problem of my code.
Maybe the way I submitted the stimulus set?
 @mschrimpf

mschrimpf:
Yes agree -- you can mark those tests as requiring private access using markers: https://github.com/brain-score/brain-score/tree/master/tests#markers
See here for an example: https://github.com/brain-score/brain-score/blob/9fbf4eda24d081c0ec7bc4d7b5572d8c13dc92d2/tests/test_benchmarks/test_freemanziemba2013.py#L8

mschrimpf:
There's a remaining error in `tests/test_stimuli.py:253`: `brainio.lookup.StimulusSetLookupError: 'Stimulus set neil.Islam2021 not found'`

mschrimpf:
I think that test also needs the private marker

mschrimpf:
I'll close and re-open this to re-trigger jenkins, looks like the job got lost

mschrimpf:
Some test failures that are unrelated to this PR. We're fixing these in #366, but please let us know if you prefer to have this PR merged more urgently @GaspTO.

GaspTO:
No worries. It's not urgent to have this PR merged.
Although, the tests seem to be passing now

PR #362, opened 2023-03-22
Jacob 2020 Behavioral Benchmarks: 3D Processing 

mike-ferguson:
This PR will contain [Jacob et al. 2020](https://www.nature.com/articles/s41467-021-22078-3#Sec20) 3D Processing Index benchmark:

Fundamentally, humans and models are performing different tasks in this benchmark. Models are passively viewing
a stimuli (and having those activations retrieved from IT), whereas humans performed a visual search task
in the papers listed below. The reasons for doing so are to replicate benchmark work done in [Jacob et al. 2020](https://www.nature.com/articles/s41467-021-22078-3#Sec20).

Following [Jacob et al. 2020](https://www.nature.com/articles/s41467-021-22078-3#Sec20), we retrieve internal neural activity from the model (layer IT)
rather than having the model perform the same visual search task that humans were performing
in the behavioral experiment from which we take the behavioral data to compare against (below):

1. Square Benchmark: [ Enns and Rensink 1991](https://www2.psych.ubc.ca/~rensink/publications/download/E&R-PsychRev-91.pdf),  Page 4, Experiment 1
2. Y Benchmark: [Enns and Rensink 1990](https://www.jstor.org/stable/40062736?seq=2), Page 324, Experiment 1

Currently, the metric is defined as "Data Cloud Comparison", which takes in means and standard errors from the reported behavioral data above, and generates more human data via a normal distribution with the above mean and SEM. Currently the metric is reporting the following averages (reproduction confirmation):

1. [Jacob et al. 2020](https://www.nature.com/articles/s41467-021-22078-3#Sec20) Reports a Square Human Baseline (from paper 1 above) of 0.76
    - Current Metric (Data Cloud Comparison) is generating an overage of 0.741, STD of 0.066
2. [Jacob et al. 2020](https://www.nature.com/articles/s41467-021-22078-3#Sec20) reports a Y Human baseline (from paper 2 above) of 0.40
    - Current Metric (Data Cloud Comparison) is generating an average of 0.419, STD of 0.059
    
 Note some variance is to be expected with metric averages, due to the normal generation process. 

The data cloud comparison metric thus then takes a single model index, say 0.12, and a vector of human indexes (generated from the cloud above), and for each human index computes a Brain-Score based on the following formula:

raw_score = max((1 - distance.euclidean(human_score, source)), 0)

The final raw score is then the average of all of the human scores. The ceiling is computed the same way, with one human index being compared against the vector of all others, for each human, and returning the average. The current square ceiling value is 0.896, and the y benchmark ceiling is currently 0.836.


deirdre-k:
- [ ]

PR #371, opened 2023-04-28
let travis run private and public tests separately

mschrimpf:
previously, when private tests were run, travis ran private _and_ public tests, and then in a separate run only the public tests. This changes it so that _only_ private and _only_ public tests are run separately.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,8 +21,8 @@ install:
 - pip install .
 - pip list # list installed package versions
 script:
-- if [ "$PRIVATE_ACCESS" = 1 ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then pytest -m "not requires_gpu and not memory_intense and not slow and not travis_slow"; fi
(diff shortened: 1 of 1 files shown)


kvfairchild (approved):

.travis.yml
```diff
@@ -21,8 +21,8 @@ install:
 - pip install .
 - pip list # list installed package versions
 script:
-- if [ "$PRIVATE_ACCESS" = 1 ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then pytest -m "not requires_gpu and not memory_intense and not slow and not travis_slow"; fi
-- if [ "$PRIVATE_ACCESS" != 1 ]; then pytest -m "not requires_gpu and not memory_intense and not private_access and not slow and not travis_slow"; fi
+- if [ "$PRIVATE_ACCESS" = 1 ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then pytest -m "private_access and not requires_gpu and not memory_intense and not slow and not travis_slow"; fi
```
kvfairchild, line 24:
```suggestion
- if [ "$PRIVATE_ACCESS" = 1 ] && [ "$TRAVIS_PULL_REQUEST" = "false" ]; then pytest -m "private_access"; fi
```

PR #377, opened 2023-07-12
THINGS: Brain-Score

linus-md:
Hello @mschrimpf,

I added the odd-one-out task to the model interface. I will write a more detailed documentation later. Is the example  looking somewhat correct?

Best,
Linus


mschrimpf (commented):
Yes good start -- I actually hadn't thought about implementing the task the way you have set it up, i.e. starting the `odd_one_out` task and then having it `look_at` all triplets of stimuli. I like it better conceptually, but am worried about practical runtime. Could you discuss with Martin Hebart what he recommends based on his experience?

The two options are:
1. iterate explicitly (specification in this PR right now)
```
candidate.start_task(BrainModel.Task.odd_one_out)
for stimulus_triplet in triplets:  # iterate over all triplets _of the experiment_
    choice = candidate.look_at(stimulus_triplet)
...
```
2. all possible triplets
```
candidate.start_task(BrainModel.Task.odd_one_out_triplet)
choices = candidate.look_at(all_stimuli)  # process everything at once, output matrix for _all possible_ triplets
```

linus-md:
Yes, we will discuss this.

brainscore/benchmarks/hebart2023.py
```diff
@@ -0,0 +1,39 @@
...
+          month = {feb},
+          pages = {e82580},
+          title = {THINGS-data, a multimodal collection of large-scale datasets for investigating object representations in human brain and behavior},
+          volume = 12,
+          year = 2023
+          }"""
+
+class _Hebart2023Accuracy(BenchmarkBase):
+    # behavioral benchmark
+    def __init__(self, dataset):
```
mschrimpf, line 17:
```suggestion
class Hebart2023Accuracy(BenchmarkBase):
    def __init__(self):
```
You can simplify this unless you are planning different versions of the same benchmark type.

linus-md:
I thought you wanted me to set up something general like this. But you are right, there would be a lot of redundant passes through the network. It would probably be better if we pass all images (1854) through the neural net, save the activations and do the odd one out task afterwards. 

Now that I think about it this would actually be incredibly inefficient. We have 470,000 validation triplets which would mean, that there would be 1,410,000 passes through the neural net rather than 1,854.

So I would propose something like this:
1. look at all images and save the activations
2. only input the activations to the odd-one-out task
3. compute score

mschrimpf:
> So I would propose something like this:
> 
>     1. look at all images and save the activations -> 1854*len(layer) feature matrix
>     2. only input the activations to the odd-one-out task -> input the three corresponding rows
>     3. compute score

I would change this to
1. tell the model to perform a triplet odd one out task
2. have the model look at the images, the model will return what is instructed to according to the specification
   a. return the odd one out for all possible triplet tasks (internally the model can compute features once and then compute all odd-one-outs)
   b. or alternatively the model returns all pairwise similarities and the benchmark uses this to approximate the triplets. (In general I'm trying to emphasize that we should ask the model for a *behavior* (like the humans), not for neural features)
3. compute score between model-predicted triplets

linus-md:
Are you aware that there are Bin(1854, 3) = 1,060,412,604 possible triples and that we only have 4,700,000 human samples? Computing all possible triples would be unnecessary and computationally expensive.

Then maybe we stick to  _iterate explicitly (specification in this PR right now)_ and memoize all activations?

mschrimpf:
That's what I was wondering about before and where Martin Hebart's experience is relevant, I'm not sure how costly this computation would be. If it's just some matrix multiplies, it might not be so bad and actually more efficient than retrieving the features one after the other. 

An improvement over the current explicit iteration where we only pass in three stimuli at a time could be to pass in all triplets at once such that all groups of three belong to one task. E.g. `model.look_at([stim1_1, stim1_2, stim1_3, stim2_1, stim2_2, stim2_3, stim3_1, ...])`

linus-md:
Martin suggest that we compute all `1,717,731` pairwise similarities and store them. He says we could iterate over the triplets afterwards or parallelize by going from triplets as `[stim_1, stim_2, stim_3]` to `[stim_1 stim_2, stim_1 stim_3, stim_2 stim_3]`. We go from a `170,000 * 3` matrix to a `170,000 * 2 * 3` matrix and call something like
```def sub2ind(array_shape, rows, cols):
        ind = np.ix_(rows, cols)
        return np.ravel_multi_index(ind, array_shape)```
which would give us all indices in the pairwise similarities matrix and we could conceivably do the odd-one-out computations parallel.

**Comment by mschrimpf** [comment]
> Martin suggest that we compute all `1,717,731` pairwise similarities

yes, great.

So in that setting the model still performs a `triplet_odd_one_out` task? Just the way it computes the odd one out is:
1. compute features for all stimuli
2. compute pairwise similarities between all pairs of stimuli
3. compute odd one out for the triplets given by the benchmark

Or does the model perform a `pairwise_similarities` task and end at 2?

**Comment by linus-md** [comment]
Is there more to do in `Hebart2023Accuracy.__call__` for now? Do we need a `load_assembly(dataset)` function like Geirhos et al.?

File: brainscore/benchmarks/hebart2023.py [hunk 2]
```diff
@@ -19,20 +20,33 @@ def __init__(self):
         self._visual_degrees = None
 
         self._number_of_trials = None
+        
+        #TODO I am not sure where this goes , but it should be in the benchmark right?
+        self.validation_data = None  # https://osf.io/b2a4j we'll need to convert this to DataAssembly right?
 
-        super(_Hebart2023Accuracy, self).__init__(
+        # TODO do we put the similarity matrix inside look_at?
```
mschrimpf, line 27: if you mean the human data, that is just kept in the assembly and then compared against the model assembly in `__call__` and more specifically the metric

brainscore/benchmarks/hebart2023.py
```diff
@@ -19,20 +20,33 @@ def __init__(self):
...
             parent=None,
             bibtex=BIBTEX)
 
     def __call__(self, candidate: BrainModel):
-        candidate.start_task(BrainModel.Task.odd_one_out, None)
-        odd_one_out = candidate.look_at(None, number_of_trials=self._number_of_trials)
-        raw_score = None
+        candidate.start_task(BrainModel.Task.odd_one_out, None) 
+        validation_data = self.validation_data
+        odd_one_outs = candidate.look_at(validation_data, number_of_trials=self._number_of_trials)  # TODO do we put the similarity matrix inside look_at?
```
mschrimpf, line 41: The candidate should look at stimuli, not the data. Typically we keep this as part of the assembly, i.e.:
```suggestion
        predicted_odd_one_outs = candidate.look_at(self._assembly.stimulus_set, number_of_trials=self._number_of_trials)
```

brainscore/benchmarks/hebart2023.py
```diff
@@ -19,20 +20,33 @@ def __init__(self):
...
+        raw_score = len(validation_data[:, 2][validation_data[:, 2]==odd_one_outs])/len(validation_data) # @Martin last entry is human odd-one-out
```
mschrimpf, line 42: I would use the [`Accuracy` metric](https://github.com/brain-score/brain-score/blob/master/brainscore/metrics/accuracy.py) here to obtain a `Score` object
```suggestion
        raw_score = self._metric(predicted_odd_one_outs, self._assembly)
```

brainscore/benchmarks/hebart2023.py
```diff
@@ -14,39 +18,32 @@
...
+        def load_assembly(dataset):                                        # 
+            pass                                                           #
+        self.stimulus_set  = brainscore.get_stimulus_set("Hebart2023")     # <- this is from Islam2021
+        
         super(Hebart2023Accuracy, self).__init__(
-            identifier=None, version=1,
-            ceiling_func=None,
-            parent=None,
+            identifier=f'Hebart2023Accuracy_{similarity_measure}', version=1,
+            ceiling_func=None,                                             # Will need the noise ceiling from Martin H.
```
mschrimpf, line 35: ideally we can compute this directly from the data to make the ceiling estimate reproducible

brainscore/benchmarks/hebart2023.py
```diff
@@ -32,7 +33,7 @@ def load_assembly(dataset):                                        #
         
         super(Hebart2023Accuracy, self).__init__(
             identifier=f'Hebart2023Accuracy_{similarity_measure}', version=1,
-            ceiling_func=None,                                             # Will need the noise ceiling from Martin H.
+            ceiling_func=0.6844,                                             # Will need the noise ceiling from Martin H.
```
mschrimpf, line 36:
```suggestion
            ceiling_func=lambda: Score([0.6844, np.nan], coords={'aggregation': ['center', 'error']}, dims=['aggregation']),
```

linus-md:
@mschrimpf @mike-ferguson what is the problem with readthedocs?

kvfairchild:
> @mschrimpf @mike-ferguson what is the problem with readthedocs?

It looks like there's a new `build.os` config key [requirement active as of today](https://blog.readthedocs.com/use-build-os-config/) to be in compliance with Read the Docs v2.0. PR #390 should fix this once merged.

linus-md:
This has all been moved here #434.

PR #383, opened 2023-08-01
use pyproject.toml instead of setup.py

mschrimpf:


mschrimpf:
The failing tests are metrics failures which are separate from this PR 
image

PR #395, opened 2023-11-06
add model conversion helpers

mschrimpf:

PR #408, opened 2023-12-11
Add models for testing model conversion helpers (based off #395)

shehadak:
(Re-opening #398)

Adds `alexnet` and `pixels` for testing.
Based off #395


tests/test_models.py
```diff
@@ -0,0 +1,15 @@
...
+# generic tests for model plugins
+# ##
+
+import pytest
+
+import brainscore_vision
+
+
+@pytest.mark.travis_slow
+@pytest.mark.parametrize('identifier',
```
kvfairchild, line 11: @mschrimpf - if we want to ensure that we can at minimum load all models (and maybe not a bad idea for other plugins as well) do we want to parametrize this with a list of ALL identifiers that we extract from the registry builder?

kvfairchild (approved):

tests/test_models.py
shehadak: I added `alexnet` in https://github.com/brain-score/brain-score/pull/408/commits/cce169123c6dbdeaf313116a76de09da26127c90. I think it is a reasonable standard to have all models in the registry tested here.

mschrimpf:
image
@shehadak @kvfairchild the tests for this PR fail for me locally. Did they pass on your end? (Neither Travis nor Jenkins ran these tests.)

shehadak:
> image @shehadak @kvfairchild the AlexNet test for this PR fails for me locally, see above. Did this pass on your end? (Neither Travis nor Jenkins ran these tests. The unit test for Pixels passes.)
> 
> I'm also not sure where the expected score of `0.6659` comes from. It should be `.508` (ceiled) / `.59` (raw), according to both [candidate models](https://github.com/brain-score/candidate_models/blob/adf86cd6d46dcfca7005e8ccb658e4e347eb79a1/tests/test___init__.py#L116) and the website.

You are correct. This is being addressed in #415.

PR #416, opened 2023-12-19
Domain-Transfer Benchmarks 

ernestoBocini:
In this PR I present two new benchmarks and some updates on Crossvalidation metrics.

Here is an overview of the main updates:

- `packaging/domain_transfer/`: contains the code to read _Oleo_ and _Pico_ assemblies. Also contains the scripts to merge them into a single merged assembly. The validity of the merge, is given by the study of the internal consistency computed across neurons that shouldn't change from the single assembly to the merged asssembly. Here is the plot that verifies that this is the case(obtainied in the` investigation_consistency.ipynb` notebook) :

output

- `brainscore/benchmarks/domain_transfer_neural.py`: new neural benchmark that uses the merged assembly and computes neural IT alignment. Here is a comparison of the trend of this benchmark on a selection of models against the MajajHong2015:
output

- `brainscore/benchmarks/domain_transfer_analysis.py`: new 'analysis' ingegneristic benchmark. Aimed at reproducing previous work ([I Gusti Bagus et al. 2022](https://dicarlolab.mit.edu/sites/default/files/publications/2022_GustiBagus_NeurIPS.pdf)), with the ability of expanding to all the models on Brain-Score. Not meant to go public. Here is a comparison proving the validity:
Screenshot 2023-12-19 at 09 01 14

- **Updates** to **cross-validation** related functions in `metrics/regression.py` & `metrics/transformations.py` : added a 'hook' to the already existing implementation of `CrossValidation`. It is now able to take among the `crossvalidation_kwargs` an extra argument, called '`preprocess_indices`' that is `None` by default and will not influence anything that doesn't specify a function for it. It's ment to provide a non invasive solution to add some further preprocessing of the indices before computing the cross-validation. `preprocess_indices` is customizable. In the case of `brainscore/benchmarks/domain_transfer_neural.py`, it is specified to be able to ensure no leakage of _background_id_ from train to test .


ernestoBocini:
New, updated image for the `domain_transfer_neural.py` neural benchmark. here plotted against all the 5 existing IT benchmarks (in case the model was scored on them)

output

ernestoBocini:
Unit tests for the new neural benchmark passing on my machine. Warnings due to old version of libraries/python. 

Screenshot 2023-12-20 at 08 29 29

brainscore/benchmarks/domain_transfer_analysis.py
```diff
@@ -0,0 +1,301 @@
...
+            crossdomain_test_images_dict, complete_training_data = split_training_test_images(crossdomain_data_dictionary=crossdomain_data_dict)
+
+            # Train the decoder #
+            decoder = get_decoder(data=complete_training_data, estimator=self._metric)
+            for crossdomain in crossdomain_test_images_dict.keys():
+                test_accuracy = get_classifier_score_2AFC(classifier=decoder, data=crossdomain_test_images_dict[crossdomain])
+                if crossdomain not in crossdomain_results:
+                    crossdomain_results[crossdomain] = [test_accuracy]
+                else:
+                    crossdomain_results[crossdomain].append([test_accuracy])
```
mschrimpf, line 84:
```suggestion
                if crossdomain not in crossdomain_results:
                    crossdomain_results[crossdomain] = []
                crossdomain_results[crossdomain].append([test_accuracy])     
```

brainscore/benchmarks/domain_transfer_analysis.py
```diff
@@ -0,0 +1,301 @@
...
+            for crossdomain in crossdomain_test_images_dict.keys():
+                test_accuracy = get_classifier_score_2AFC(classifier=decoder, data=crossdomain_test_images_dict[crossdomain])
+                if crossdomain not in crossdomain_results:
+                    crossdomain_results[crossdomain] = [test_accuracy]
+                else:
+                    crossdomain_results[crossdomain].append([test_accuracy])     
+
+        for key, values in crossdomain_results.items():
+
+            crossdomain_results[key] = [np.mean(values)[0], np.std(values)[0]]
```
mschrimpf, line 88: the `[0]` is confusing -- `np.mean` and `np.std` should return a scalar already?

brainscore/benchmarks/domain_transfer_analysis.py
```diff
@@ -0,0 +1,301 @@
...
+
+        with open('crossdomain_data_dict.pkl', 'wb') as file:
+            pickle.dump(crossdomain_data_dict, file)
+        with open('crossdomain_results_dict.pkl', 'wb') as file:
+            pickle.dump(crossdomain_results, file)
```
mschrimpf, line 93: what are these for?

brainscore/benchmarks/domain_transfer_analysis.py
```diff
@@ -0,0 +1,301 @@
...
+            
+        return crossdomain_results
+    
+
+def OOD_AnalysisBenchmark():
+    return _OOD_AnalysisBenchmark(
+        metric=RidgeClassifierCV(alphas=[0.0001, 0.001, 0.01, 0.1, 1, 10], fit_intercept=True, normalize=True)
```
mschrimpf, line 100: this is not a metric but a classifier. The metric would be accuracy (https://github.com/brain-score/brain-score/blob/master/brainscore/metrics/accuracy.py)

brainscore/benchmarks/domain_transfer_neural.py
```diff
@@ -0,0 +1,137 @@
+# import general libraries
+import os
+
+# import brain-score specific libraries
+from brainscore.benchmarks._neural_common import NeuralBenchmark, average_repetition
+from brainscore.metrics.ceiling import InternalConsistency
+from brainscore.metrics.regression import CrossRegressedCorrelation, pls_regression, pearsonr_correlation # OODRegressedCorrelation, CrossRegressedCorrelation_with_Background
```
mschrimpf, line 7: dead code
```suggestion
from brainscore.metrics.regression import CrossRegressedCorrelation, pls_regression, pearsonr_correlation
```

brainscore/benchmarks/domain_transfer_neural.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+
+def IT_pls():
+    return _commonbase('IT', identifier_metric_suffix='pls',
+                                       similarity_metric=CrossRegressedCorrelation(
+                                           regression=pls_regression(), correlation=pearsonr_correlation(),
+                                           crossvalidation_kwargs = {'stratification_coord': 'object_label',
+                                                                     'preprocess_indices': _preprocess_indices}), 
+                                       ceiler=InternalConsistency())
```
mschrimpf, line 55: can just combine, it's only IT data and we're only using one metric right?
```suggestion
def Igustibagus2024():
    assembly_repetition = LazyLoad(lambda: load_domain_transfer(average_repetitions=False))
    assembly = LazyLoad(lambda: load_domain_transfer(average_repetitions=True))
    similarity_metric=CrossRegressedCorrelation(
                                           regression=ridge_regression(), correlation=pearsonr_correlation(),
                                           crossvalidation_kwargs = {'stratification_coord': 'object_label',
                                                                     'preprocess_indices': _preprocess_indices})
    ceiler=InternalConsistency()
    return NeuralBenchmark(identifier=f'Igustibagus2024-pls', version=1, parent='IT',
                        assembly=assembly, similarity_metric=similarity_metric,
                        visual_degrees=VISUAL_DEGREES, number_of_trials=NUMBER_OF_TRIALS,
                        ceiling_func=lambda: ceiler(assembly_repetition),
                        bibtex=BIBTEX)
```
although maybe we want to do ridge instead of pls?

brainscore/benchmarks/domain_transfer_neural.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+#### helpers ####
+
+def load_domain_transfer(average_repetitions, region):
```
mschrimpf, line 60: most of this function should be replaced by `brainscore.get_assembly('Igustibagus2024')` right? Once you packaged and uploaded it.

brainscore/benchmarks/domain_transfer_neural.py
```diff
@@ -0,0 +1,137 @@
...
+
+        common_non_zero = train_set.intersection(test_set) - {0}
+        if not common_non_zero:
+            return train_ids, test_ids
+
+        train_indices_to_retain = [i for i, bg_id in enumerate(train_ids) if bg_id not in common_non_zero]
+
+        return train_indices_to_retain
+
+def _preprocess_indices(train_indices, test_indices, source_assembly):
```
mschrimpf, line 129: these two functions need comments what they achieve and why

brainscore/metrics/transformations.py
```diff
@@ -170,11 +170,13 @@ class Defaults:
         stratification_coord = 'object_name'  # cross-validation across images, balancing objects
         unique_split_values = False
         random_state = 1
+        preprocess_indices = None
```
mschrimpf, line 173:
```suggestion
```

brainscore/metrics/transformations.py
```diff
@@ -170,11 +170,13 @@ class Defaults:
...
 
     def __init__(self,
                  splits=Defaults.splits, train_size=None, test_size=None,
                  split_coord=Defaults.split_coord, stratification_coord=Defaults.stratification_coord, kfold=False,
-                 unique_split_values=Defaults.unique_split_values, random_state=Defaults.random_state):
+                 unique_split_values=Defaults.unique_split_values, random_state=Defaults.random_state,
+                 preprocess_indices=Defaults.preprocess_indices):
```
mschrimpf, line 179:
```suggestion
                 preprocess_indices=None):
```

brainscore/metrics/transformations.py
```diff
@@ -311,6 +317,13 @@ def pipe(self, source_assembly, target_assembly):
         split_scores = []
         for split_iterator, (train_indices, test_indices), done \
                 in tqdm(enumerate_done(splits), total=len(splits), desc='cross-validation'):
+            
+            if hasattr(self, '_preprocess_indices'):
+                if self._preprocess_indices is not None:
+                    print('train_indices before: ', len(train_indices))
+                    train_indices, test_indices = self._preprocess_indices(train_indices, test_indices, source_assembly)
+                    print('train_indices after: ', len(train_indices))
```
mschrimpf, line 325:
```suggestion
            if self._preprocess_indices is not None:
                train_indices, test_indices = self._preprocess_indices(train_indices, test_indices, source_assembly)
```
no prints. If these are important to log, use a `logger`.

packaging/domain-transfer/merged_assembly/create_stimulus_set_with_background-id.py
```diff
@@ -0,0 +1,34 @@
...
+from helpers_background_id import loading_brain_model_activation
+
+
+# get the background-id
+brain_model_name='dcgan'
+image_source='dicarlo.domain_transfer'
+
+domain_transfer_data = loading_brain_model_activation(brain_model_name, image_source, penultimate_layer=False)
+
+background_df = pd.DataFrame({'id': domain_transfer_data.stimulus_id.values, 'background': domain_transfer_data.background_id.values})
```
mschrimpf, line 18: you don't need a model to achieve this, the `background_id` is in the stimulus set

tests/test_benchmarks/test_domain_transfer_neural.py
```diff
@@ -0,0 +1,20 @@
...
+
+@pytest.mark.memory_intense
+@pytest.mark.private_access
+class TestAssembly:
+    def test_IT(self):
+        assembly = load_domain_transfer(average_repetitions=True, region='IT')
+        check_standard_format(assembly)
+        assert assembly.attrs['stimulus_set_identifier'] == 'sanghavi_dicarlo_stimulus_set_domain_transfer'
+        assert set(assembly['region'].values) == {'IT'}
+        assert len(assembly['presentation']) == 780
```
mschrimpf, line 15:
```suggestion
        assert len(assembly['presentation']) == 780
        assert set(assembly['domain']) == {'hvm', 'convex_hull', 'outline', 'silhouette', 'cartoon', 'line_drawing', 'mosaic', 'painting', 'sketch'}
```

brainscore/benchmarks/domain_transfer_analysis.py
ernestoBocini: they return `array([scalar])`

brainscore/benchmarks/domain_transfer_analysis.py
ernestoBocini: to produce stor results and produce the plots I included in this PR. I remove them

brainscore/benchmarks/domain_transfer_neural.py
ernestoBocini: I was thinking about keeping them separated, to make it easy to add new metrics , e.g. ridge as you mentioned

brainscore/benchmarks/domain_transfer_neural.py
ernestoBocini: Not really, because the assembly version I uploaded contains all the stimuli and is not filtered for consistency > 0.7  . That's what we agreed on, right?

packaging/domain-transfer/merged_assembly/create_stimulus_set_with_background-id.py
ernestoBocini: That would be right if I were to define the background indices from scratch. However, for the sake of time, I used Marlia's code, in which she defined `background_id` as a `presentation` dimension of the results of the `look_at` function, which needs a model (any model).

brainscore/metrics/transformations.py
ernestoBocini: Removing it from `Defaults` means that I need to call an instance of Splits() when accessing it in the `CrossValidation(Transformation)` like this:
```
class CrossValidation(Transformation):
    def __init__(self, *args, split_coord=Split.Defaults.split_coord,
                 stratification_coord=Split.Defaults.stratification_coord,
                 preprocess_indices=Split()._preprocess_indices, **kwargs):
        self._split_coord = split_coord
        self._stratification_coord = stratification_coord
        self._split = Split(*args, split_coord=split_coord, stratification_coord=stratification_coord, **kwargs)
        self._logger = logging.getLogger(fullname(self))
        self._preprocess_indices = preprocess_indices
```

I wonder whether it would be better to leave it in `Defaults` maybe?

packaging/domain-transfer/merged_assembly/create_stimulus_set_with_background-id.py
mschrimpf: the `presentation` coordinates in `look_at` come exclusively from the stimulus set, so the background id needs to already be present there

brainscore/benchmarks/domain_transfer_neural.py
mschrimpf: let's separate them when we need to ("Premature optimization is the root of all evil")

brainscore/benchmarks/domain_transfer_neural.py
mschrimpf: I guess I'm saying that you shouldn't need 40 lines of code to load the assembly -- most of this method seems like it can be simplified substantially

mschrimpf (approved):
tests pass for me locally -- jenkins tests are currently busted with the 2.0 refactor, so will merge

PR #439, opened 2024-01-05
README.md

linus-md:
Shouldn't ``pip install git+https://github.com/brain-score/brain-score`` be ``pip install git+https://github.com/brain-score/vision`` given that the repo name was changed and that ``https://github.com/brain-score/brain-score`` redirects to ``https://github.com/brain-score/vision``?


linus-md:
Never mind I just saw #432.

PR #492, opened 2024-01-10
Add microsaccades to Brain-Score core

benlonnqvist:
In this PR we add microsaccades to Brain-Score core. Integration of [model_tools PR #75](https://github.com/brain-score/model-tools/pull/75) to Brain-Score 2.0.

**Conceptual Problem**: We need variance in model responses to static stimuli.

**Conceptual Solution**: Microsaccades are a prevalent phenomenon in vision and visual experiments. They are involuntary movements of the eye at very small scales. Microsaccades are not (and often can not due to their small extent) be controlled for in experiments, and are thus implicitly baked into the variance in primate responses. As such, microsaccades are a phenomenon that is either explicitly or implicitly believed to be unimportant for behavioral/neural outcomes, and thus a good candidate for getting variance in responses to static stimuli. (e.g. Rolfs 2009, Vis. Res.; Haddad & Steinmann 1973, Vis. Res.)

**Technical problem**: The current Brain-Score interface does not support microsaccades.

**Technical solution**: the [ActivationsExtractorHelper](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/core.py#L22) will be changed. The [from_paths](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/core.py#L59) method now takes a path, loads the image from the path, does pixel position jittering to the image, and saves the jittered image into a temporary new path. The new path is passed to the downstream functions to get the model activations.

To accomplish this, the from_paths method will now receive a dictionary of tags. This dictionary of tags comes from a call to the [look_at](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/brain_transformation/neural.py#L26) method, where number_of_trials is replaced with a new dictionary. This new dictionary then traverses the following path:

- [run_activations](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/brain_transformation/neural.py#L39)
- A call to [activations_model](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/brain_transformation/neural.py#L39)
- In practice, this calls a [TensorFlowWrapper](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/tensorflow.py#L6)/[PytorchWrapper](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/pytorch.py#L13)/etc.’s [self._extractor](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/pytorch.py#L22) class’s \_\_call\_\_() method in its \_\_call\_\_() method. (I.e., [PytorchWrapper.\_\_call\_\_()](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/pytorch.py#L40) => [self._extractor.\_\_call\_\_()](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/pytorch.py#L41) )
- This extractor’s call method is [ActivationsExtractorHelper.\_\_call\_\_()](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/core.py#L36)
- Thus we pass the new dictionary to [ActivationsExtractorHelper.\_\_call\_\_()](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/activations/core.py#L36) where it will be parsed.

TODO list:

- [x] Implement microsaccades
- [x] Implement unit tests for microsaccades

Other Brain-Score benchmark PRs depend on this PR (https://github.com/brain-score/vision/pull/365).

Huge thanks to Hannes Mehrer (@hanme) who implemented microsaccades, which I've only adapted for the generalized case here.


benlonnqvist:
It seems that the Jenkins build is skipping tests on this PR, which is not the intended case: ```http://braintree.mit.edu:8080/job/unittest_brainscore/1516/parsed_console/``` shows the log:
``` 
fatal: ambiguous argument 'origin/vision_microsaccades': unknown revision or path not in the working tree.
Use '--' to separate paths from revisions, like this:
'git <command> [<revision>...] -- [<file>...]'
...
MODIFIES_PLUGIN: False
PLUGIN_ONLY: True
Changed files:  
Only plugins modified. Will skip tests.
```

Unsure as to why this happens, and whether it's something I can fix as a part of the PR? The intended case certainly is for tests to run on this PR. Any help would be much appreciated!

kvfairchild:
Hi @benlonnqvist - thanks for flagging this; it looks like you've caught an issue with the way we're handling forked PRs in Jenkins. This should raise an assertion error in the Travis PR build, but that isn't able to run while there are unresolved conflicts with `master`. I should be able to push a fix later today and will let you know!

benlonnqvist:
Thank you @kvfairchild! Much appreciated.

kvfairchild:
Hi @benlonnqvist - can you try again now?

benlonnqvist:
Hi @kvfairchild, tests seem to be working now, thanks again!

benlonnqvist:
Hi! I think there are some remaining tests that fail on Jenkins that I don't think have anything to do with the PR itself - although I could be wrong. Perhaps it might be possible to have a quick look as to what might be going wrong with the few remaining tests and whether they have anything to do with this PR specifically? Thanks!

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +36,41 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
         self._stimulus_set_hooks = {}
         self._batch_activations_hooks = {}
 
-    def __call__(self, stimuli, layers, stimuli_identifier=None):
+    def __call__(self, stimuli, layers, stimuli_identifier=None, model_requirements=None):
```
mschrimpf, line 39: I don't like that this exposes the micro-saccades to the benchmark. It's a modeling choice of if/how to do micro-saccades. E.g. models with micro-saccades would yield different responses over multiple trials of the same stimulus (we already have `number_of_trials`).

benlonnqvist:
Hi, it seems that Jenkins is now crashing on building the PR altogether (see [here](http://braintree.mit.edu:8080/job/unittest_brainscore/1621/console)). I don't think this is related to the PR - is there something I can do to resolve the issue potentially?

mschrimpf:
re-trigger jenkins

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -150,6 +248,23 @@ def _get_batch_activations(self, inputs, layer_names, batch_size):
         activations = self._unpad(activations, num_padding)
         return activations
 
+    def translate_images(self, images, shift):
+        assert type(images) == list
```
mschrimpf, line 252:
```suggestion
        assert isinstance(images, list)
```
but even with that change, this is not very pythonic. Why does the type matter? Just do duck typing

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -209,28 +336,75 @@ def _package_layer(self, layer_activations, layer, stimuli_paths):
...
-        coords = {'stimulus_path': stimuli_paths,
-                  'neuroid_num': ('neuroid', list(range(activations.shape[1]))),
-                  'model': ('neuroid', [self.identifier] * activations.shape[1]),
-                  'layer': ('neuroid', [layer] * activations.shape[1]),
-                  }
+        if require_variance:
+            coords = self.build_microsaccade_coords(activations, layer, stimuli_paths)
+        else:
+            coords = {'stimulus_path': ('presentation', stimuli_paths),
+                      'stimulus_path2': ('presentation', stimuli_paths),  # to avoid DataAssembly dim collapse
```
mschrimpf, line 343: I get why you're doing this, but is there no other useful information for `presentation` to include rather than duplicating this coord? 😅

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -209,28 +336,75 @@ def _package_layer(self, layer_activations, layer, stimuli_paths):
...
+                      'neuroid_num': ('neuroid', list(range(activations.shape[1]))),
+                      'model': ('neuroid', [self.identifier] * activations.shape[1]),
+                      'layer': ('neuroid', [layer] * activations.shape[1]),
+                      }
```
mschrimpf, line 347:
```suggestion
        coords = {'stimulus_path': ('presentation', stimuli_paths),
                  'stimulus_path2': ('presentation', stimuli_paths),  # to avoid DataAssembly dim collapse
                  'neuroid_num': ('neuroid', list(range(activations.shape[1]))),
                  'model': ('neuroid', [self.identifier] * activations.shape[1]),
                  'layer': ('neuroid', [layer] * activations.shape[1]),
              }
          if require_variance:
              microsaccade_coords = self.build_microsaccade_coords(activations, layer, stimuli_paths)
              coords = {**coords, **microsaccade_coords}
            
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -209,28 +336,75 @@ def _package_layer(self, layer_activations, layer, stimuli_paths):
...
 
+    def build_microsaccade_coords(self, activations, layer, stimuli_paths):
+        coords = {
+            'stimulus_path': ('presentation', stimuli_paths),
+            'shift_x': ('presentation', [shift[0] for shift in self.shifts]),
+            'shift_y': ('presentation', [shift[1] for shift in self.shifts]),
+            'neuroid_num': ('neuroid', list(range(activations.shape[1]))),
+            'model': ('neuroid', [self.identifier] * activations.shape[1]),
+            'layer': ('neuroid', [layer] * activations.shape[1]),
+        }
```
mschrimpf, line 368:
```suggestion
        coords = {
            'shift_x': ('presentation', [shift[0] for shift in self.shifts]),
            'shift_y': ('presentation', [shift[1] for shift in self.shifts]),
        }
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -265,15 +439,62 @@ def attach_stimulus_set_meta(assembly, stimulus_set):
     stimulus_paths = [str(stimulus_set.get_stimulus(stimulus_id)) for stimulus_id in stimulus_set['stimulus_id']]
     stimulus_paths = [lstrip_local(path) for path in stimulus_paths]
     assembly_paths = [lstrip_local(path) for path in assembly['stimulus_path'].values]
+
     assert (np.array(assembly_paths) == np.array(stimulus_paths)).all()
-    assembly['stimulus_path'] = stimulus_set['stimulus_id'].values
+    assembly = assembly.reset_index('presentation')
+    assembly.assign_coords(stimulus_path=('presentation', stimulus_set['stimulus_id'].values))
```
mschrimpf, line 445: is this different from
```suggestion
    assembly['stimulus_path'] = 'presentation', stimulus_set['stimulus_id'].values)
```
?

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -265,15 +439,62 @@ def attach_stimulus_set_meta(assembly, stimulus_set):
...
     for column in stimulus_set.columns:
-        assembly[column] = 'stimulus_id', stimulus_set[column].values
-    assembly = assembly.stack(presentation=('stimulus_id',))
+        assembly = assembly.assign_coords({column: ('presentation', stimulus_set[column].values)})
+        all_columns.append(column)
+    if 'stimulus_id' in all_columns:
+        all_columns.remove('stimulus_id')
+    if 'stimulus_path2' in all_columns:
+        all_columns.remove('stimulus_path2')
+    assembly = assembly.set_index(presentation=['stimulus_id', 'stimulus_path2'] + all_columns)
```
mschrimpf, line 457: I'm confused what this does. I guess this somehow tries to work around xarray multi-indexing? Needs at least a code comment, maybe can be simplified

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -265,15 +439,62 @@ def attach_stimulus_set_meta(assembly, stimulus_set):
...
     return assembly
 
 
+def attach_stimulus_set_meta_with_microsaccades(assembly, stimulus_set, number_of_trials, shifts):
```
mschrimpf, line 461: yuck lots of duplicate code again. Let's not repeat everything

brainscore_vision/model_helpers/brain_transformation/neural.py
```diff
@@ -23,21 +23,47 @@ def __init__(self, identifier, activations_model, region_layer_map, visual_degre
     def identifier(self):
         return self._identifier
 
-    def look_at(self, stimuli, number_of_trials=1):
+    def look_at(self, stimuli, number_of_trials=1, require_variance: bool = False):
+        """
+        :param number_of_trials: An integer that determines how many repetitions of the same model performs.
+        :param require_variance: A bool that asks models to output different responses to the same stimuli (i.e.,
```
mschrimpf, line 29: near duplicate comment from the ActivationsHelper. Just link to those comments.

tests/test_model_helpers/activations/test___init__.py
```diff
@@ -189,6 +189,58 @@ def test_from_image_path(model_ctr, layers, image_name, pca_components, logits):
     return activations
 
 
+@pytest.mark.parametrize("image_name", ['rgb.jpg', 'grayscale.png', 'grayscale2.jpg', 'grayscale_alpha.png',
+                                        'palletized.png'])
+@pytest.mark.parametrize(["model_ctr", "layers"], models_layers)
+@pytest.mark.parametrize("number_of_trials", [1, 5, 25])
+def test_microsaccades_from_image_path(model_ctr, layers, image_name, number_of_trials):
```
mschrimpf, line 196:
```suggestion
def test_require_variance_has_shift_coords(model_ctr, layers, image_name, number_of_trials):
```

tests/test_model_helpers/activations/test___init__.py
```diff
@@ -189,6 +189,58 @@ def test_from_image_path(model_ctr, layers, image_name, pca_components, logits):
...
+@pytest.mark.parametrize(["model_ctr", "layers"], models_layers)
+@pytest.mark.parametrize("require_variance", [False, True])
+def test_model_requirements(model_ctr, layers, image_name, require_variance):
+    stimulus_paths = [os.path.join(os.path.dirname(__file__), image_name)]
+    activations_extractor = model_ctr()
+
+    activations_with_req = activations_extractor(stimuli=stimulus_paths, layers=layers,
+                                                 number_of_trials=2, require_variance=require_variance)
+    activations_without_req = activations_extractor(stimuli=stimulus_paths, layers=layers,
+                                                    number_of_trials=1, require_variance=False)
```
mschrimpf, line 219: confused why we parametrize by `require_variance` but then also have to `activations_extractor` calls

mschrimpf (commented):
in addition to in-code comments, please delete `brainscore_vision/data/scialom2024/__init__.py`

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -30,19 +34,48 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
         :param stimuli_identifier: a stimuli identifier for the stored results file. False to disable saving.
+        :param number_of_trials: An integer that determines how many repetitions of the same model performs.
+        :param require_variance: A bool that asks models to output different responses to the same stimuli (i.e.,
+            allows stochastic responses to identical stimuli, even in deterministic models). The current implementation
+            implements this using microsaccades.
+            Human microsaccade amplitude varies by who you ask, an estimate might be <0.1 deg = 360 arcsec = 6arcmin.
+            The goal of microsaccades is to obtain multiple different neural activities to the same input stimulus
+            from non-stochastic models. This is to improve estimates of e.g. psychophysical functions, but also other
+            things. Note that microsaccades are also applied to stochastic models to make them comparable within-
+            benchmark to non-stochastic models.
```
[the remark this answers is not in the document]
benlonnqvist: Will be changed in a future commit

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -30,19 +34,48 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
+            In the current implementation, if `require_variance=True`, the model selects microsaccades according to
+            its own microsaccade behavior (if it has implemented it), or with the base behavior of saccading in
+            input pixel space with 1-pixel increments from the center of the stimulus. The base behavior thus
+            maintains a fixed microsaccade distance as measured in visual angle, regardless of the model's visual angle.
```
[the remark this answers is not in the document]
benlonnqvist: Agreed to change this computation to be explicitly based on visual angle

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -142,6 +195,51 @@ def _get_activations_batched(self, paths, layers, batch_size):
...
+            # compute activations on the entire batch one shift at a time
+            for shift in self.shifts:
+                assert type(shift) == tuple
+                temp_file_paths = self.translate_images(batch_inputs, shift)
+                activations = self._get_batch_activations(temp_file_paths, layers, batch_size)
+                for temp_file_path in temp_file_paths:
+                    try:
+                        os.remove(temp_file_path)
+                    except FileNotFoundError:
+                        pass
```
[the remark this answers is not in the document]
benlonnqvist: Agreed to keep mostly as-is (to satisfy TF models and remain explicit), but refactored into a dedicated function in a future commit

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: Not in this case, but used this to keep it uniform with the multi-coord assignment below

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +37,56 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
         self._stimulus_set_hooks = {}
         self._batch_activations_hooks = {}
 
-    def __call__(self, stimuli, layers, stimuli_identifier=None):
+        self.number_of_trials = 1  # for use with microsaccades.
+        self.microsaccade_extent_degrees = 0.05  # how many degrees models microsaccade by default
+
+        # a dict that contains the image paths and their respective microsaccades. e.g.,
+        #  {'abc.jpg': [(0, 0), (1.5, 2)]}
+        self.microsaccades = {'pixels': {}, 'degrees': {}}
```
mschrimpf, line 45: I don't understand how the comment and the dict instantiation fit together. `'pixels'` and `'degrees'` are not image paths?

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +37,56 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
+        self._visual_degrees = None  # Model visual degrees. Used for computing microsaccades
```
mschrimpf, line 46:
```suggestion
        self._visual_degrees = None  # Model visual degrees. Used for computing microsaccades in the space of degrees rather than pixels
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +37,56 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
+                 require_variance: bool = False):
         """
         :param stimuli_identifier: a stimuli identifier for the stored results file. False to disable saving.
+        :param number_of_trials: An integer that determines how many repetitions of the same model performs.
+        :param require_variance: A bool that asks models to output different responses to the same stimuli (i.e.,
+            allows stochastic responses to identical stimuli, even in otherwise deterministic base models). 
+            We here implement this using microsaccades.
+            Human microsaccade amplitude varies by who you ask, an estimate might be <0.1 deg = 360 arcsec = 6arcmin.
+            Our motivation to make use of such microsaccades is to obtain multiple different neural activities to the same input stimulus
+            from non-stochastic models. This is to improve estimates of e.g. psychophysical functions.
```
mschrimpf, line 58:
```suggestion
            from non-stochastic models. This enables models to engage on e.g. psychophysical functions which often require variance for the same stimulus.
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +37,56 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
+            In the current implementation, if `require_variance=True`, the model microsaccades in
+            input pixel space with 1-pixel increments from the center of the stimulus. The base behavior thus
```
mschrimpf, line 60: I thought we're doing this in visual degrees?

mschrimpf: same comment was below, but doesn't look like it has been addressed.

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -33,16 +37,56 @@ def __init__(self, get_activations, preprocessing, identifier=False, batch_size=
...
         if isinstance(stimuli, StimulusSet):
-            return self.from_stimulus_set(stimulus_set=stimuli, layers=layers, stimuli_identifier=stimuli_identifier)
+            return self.from_stimulus_set(stimulus_set=stimuli, layers=layers, stimuli_identifier=stimuli_identifier,
+                                          require_variance=require_variance)
         else:
-            return self.from_paths(stimuli_paths=stimuli, layers=layers, stimuli_identifier=stimuli_identifier)
+            return self.from_paths(stimuli_paths=stimuli,
+                                   layers=layers,
+                                   stimuli_identifier=stimuli_identifier,
+                                   require_variance=require_variance)
```
mschrimpf, line 87: maybe the following would be simpler?
```
function_call = functools.partial(self.from_stimulus_set, stimulus_set=stimuli) 
    if isinstance(stimuli, StimulusSet) else functools.partial(self.from_paths(stimuli_paths=stimuli)
return function_call(
    layers=layers, 
    stimuli_identifier=stimuli_identifier, 
    require_variance=require_variance)

File: brainscore_vision/model_helpers/activations/core.py [hunk 20]
```diff
@@ -53,38 +97,46 @@ def from_stimulus_set(self, stimulus_set, layers, stimuli_identifier=None):
...
-        # In case stimuli paths are duplicates (e.g. multiple trials), we first reduce them to only the paths that need
-        # to be run individually, compute activations for those, and then expand the activations to all paths again.
-        # This is done here, before storing, so that we only store the reduced activations.
-        reduced_paths = self._reduce_paths(stimuli_paths)
-        activations = fnc(layers=layers, stimuli_paths=reduced_paths)
-        activations = self._expand_paths(activations, original_paths=stimuli_paths)
+        if require_variance:
+            activations = fnc(layers=layers, stimuli_paths=stimuli_paths, require_variance=require_variance)
+        else:
+            # In case stimuli paths are duplicates (e.g. multiple trials), we first reduce them to only the paths that need
```
mschrimpf, line 119:
```suggestion
            # When we are not asked for varying responses but receive `stimuli_paths` duplicates (e.g. multiple trials), we first reduce them to only the paths that need
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -53,38 +97,46 @@ def from_stimulus_set(self, stimulus_set, layers, stimuli_identifier=None):
...
+            reduced_paths = self._reduce_paths(stimuli_paths)
+            activations = fnc(layers=layers, stimuli_paths=reduced_paths, require_variance=require_variance)
+            activations = self._expand_paths(activations, original_paths=stimuli_paths)
         return activations
 
     @store_xarray(identifier_ignore=['stimuli_paths', 'layers'], combine_fields={'layers': 'layer'})
-    def _from_paths_stored(self, identifier, layers, stimuli_identifier, stimuli_paths):
+    def _from_paths_stored(self, identifier, layers, stimuli_identifier,
+                           stimuli_paths, number_of_trials: int = 1, require_variance: bool = False):
         return self._from_paths(layers=layers, stimuli_paths=stimuli_paths)
```
mschrimpf, line 130:
```suggestion
        return self._from_paths(layers=layers, stimuli_paths=stimuli_paths, require_variance=require_variance)
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -53,38 +97,46 @@ def from_stimulus_set(self, stimulus_set, layers, stimuli_identifier=None):
...
+    def _from_paths(self, layers, stimuli_paths, require_variance: bool = False):
         if len(layers) == 0:
             raise ValueError("No layers passed to retrieve activations from")
         self._logger.info('Running stimuli')
-        layer_activations = self._get_activations_batched(stimuli_paths, layers=layers, batch_size=self._batch_size)
+        layer_activations = self._get_activations_batched(stimuli_paths, layers=layers, batch_size=self._batch_size,
+                                                          require_variance=require_variance)
         self._logger.info('Packaging into assembly')
-        return self._package(layer_activations, stimuli_paths)
+        return self._package(layer_activations, stimuli_paths, require_variance)
```
mschrimpf, line 139: whenever there's more than 1-2 arguments, I prefer to use keywords to avoid confusion and inadvertent mis-ordering.
```suggestion
        return self._package(layer_activations=layer_activations, stimuli_paths=stimuli_paths, require_variance=require_variance)
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
         self._stimulus_set_hooks[handle.id] = hook
         return handle
 
-    def _get_activations_batched(self, paths, layers, batch_size):
-        layer_activations = None
+    def _get_activations_batched(self, paths, layers, batch_size, require_variance: bool):
```
mschrimpf, line 180:
```suggestion
    def _get_activations_batched(self, paths, layers, batch_size: int, require_variance: bool):
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
         for batch_start in tqdm(range(0, len(paths), batch_size), unit_scale=batch_size, desc="activations"):
             batch_end = min(batch_start + batch_size, len(paths))
             batch_inputs = paths[batch_start:batch_end]
-            batch_activations = self._get_batch_activations(batch_inputs, layer_names=layers, batch_size=batch_size)
-            for hook in self._batch_activations_hooks.copy().values():  # copy to avoid handle re-enabling messing with the loop
+
+            batch_activations = OrderedDict()
+            # compute activations on the entire batch one shift at a time
+            for shift_number in range(self.number_of_trials):
+
```
mschrimpf, line 189:
```suggestion
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+    def _get_activations_batched(self, paths, layers, batch_size, require_variance: bool):
+        layer_activations = OrderedDict()
         for batch_start in tqdm(range(0, len(paths), batch_size), unit_scale=batch_size, desc="activations"):
             batch_end = min(batch_start + batch_size, len(paths))
             batch_inputs = paths[batch_start:batch_end]
-            batch_activations = self._get_batch_activations(batch_inputs, layer_names=layers, batch_size=batch_size)
-            for hook in self._batch_activations_hooks.copy().values():  # copy to avoid handle re-enabling messing with the loop
+
+            batch_activations = OrderedDict()
+            # compute activations on the entire batch one shift at a time
```
mschrimpf, line 187: what is a shift? explain more

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
 
-        return layer_activations
+        # fast concat all batches
+        for layer_name, layer_outputs in layer_activations.items():
+            layer_activations[layer_name] = np.concatenate(layer_outputs)
 
-    def _get_batch_activations(self, inputs, layer_names, batch_size):
+        return layer_activations  # this is all batches
+
+    def _get_batch_activations(self, inputs, layer_names, batch_size, require_variance: bool = False,
```
mschrimpf, line 213:
```suggestion
    def _get_batch_activations(self, inputs, layer_names, batch_size: int, require_variance: bool = False,
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
             batch_inputs = paths[batch_start:batch_end]
-            batch_activations = self._get_batch_activations(batch_inputs, layer_names=layers, batch_size=batch_size)
-            for hook in self._batch_activations_hooks.copy().values():  # copy to avoid handle re-enabling messing with the loop
+
+            batch_activations = OrderedDict()
+            # compute activations on the entire batch one shift at a time
+            for shift_number in range(self.number_of_trials):
+
+                activations = self._get_batch_activations(batch_inputs, layers, batch_size, require_variance,
+                                                          shift_number)
```
mschrimpf, line 191:
```suggestion
                activations = self._get_batch_activations(inputs=batch_inputs, layer_names=layers, batch_size=batch_size, require_variance=require_variance,
                                                          trial_number=shift_number)
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+
+                for layer_name, layer_output in activations.items():
+                    batch_activations.setdefault(layer_name, []).append(layer_output)
+
+            # concatenate all shifts into this batch
```
mschrimpf, line 196:
```suggestion
            # concatenate all shifts in this batch
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
-                layer_activations = copy.copy(batch_activations)
-            else:
-                for layer_name, layer_output in batch_activations.items():
-                    layer_activations[layer_name] = np.concatenate((layer_activations[layer_name], layer_output))
+            # add this batch to layer_activations
+            for layer_name, layer_output in batch_activations.items():
+                layer_activations.setdefault(layer_name, []).append(layer_output)
 
-        return layer_activations
+        # fast concat all batches
```
mschrimpf, line 207: what does the "fast" refer to?

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+            layer_activations[layer_name] = np.concatenate(layer_outputs)
 
-    def _get_batch_activations(self, inputs, layer_names, batch_size):
+        return layer_activations  # this is all batches
+
+    def _get_batch_activations(self, inputs, layer_names, batch_size, require_variance: bool = False,
+                               trial_number: int = 1):
         inputs, num_padding = self._pad(inputs, batch_size)
         preprocessed_inputs = self.preprocess(inputs)
+        preprocessed_inputs = self.translate_images(preprocessed_inputs, inputs, trial_number, require_variance)
```
mschrimpf, line 217: would prefer kwargs `param=value` over positional args

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
         activations = self.get_activations(preprocessed_inputs, layer_names)
         assert isinstance(activations, OrderedDict)
         activations = self._unpad(activations, num_padding)
+        if require_variance:
+            self.remove_temporary_files(preprocessed_inputs)
         return activations
 
+    def set_visual_degrees(self, visual_degrees: float) -> None:
```
mschrimpf, line 225:
```suggestion
    def set_visual_degrees(self, visual_degrees: float):
```
usually we wouldn't write the `None` for a function not returning anything

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+        """
+        A method used by ModelCommitments to give the ActivationsExtractorHelper their visual degrees for
+        performing microsaccades.
+        """
+        self._visual_degrees = visual_degrees
+
+    def translate_images(self, images: List[Union[str, np.ndarray]], image_paths: List[str], trial_number: int,
+                         require_variance: bool) -> List[str]:
+        """A method that translates images according to selected microsaccades, if microsaccades are required."""
```
mschrimpf, line 234:
```suggestion
        """Translate images according to selected microsaccades, if microsaccades are required."""
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+        output_images = []
+        for index, image_path in enumerate(image_paths):
+            # When microsaccades are not used, skip computing them and return the base images.
+            #  This iteration could be entirely skipped, but recording microsaccades for all images regardless
+            #  of whether variance is required or not is convenient for adding an extra presentation dimension
+            #  in the layer assembly later to avoid collapse, or otherwise extraneous mock dims.
```
mschrimpf, line 240: and to be a good citizen and keep track of metadata as much as we can

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+            #  The method could further be streamlined by calling `self.get_image_with_shape()` and
+            #  `self.select_microsaccade` for all images regardless of require_variance, but it seems like a bad
+            #  idea to introduce cv2 image loading for all models and images, regardless of whether they are actually
+            #  microsaccading.
+            if not require_variance:
+                self.microsaccades['pixels'][image_path] = [(0., 0.)]
+                self.microsaccades['degrees'][image_path] = [(0., 0.)]
+                output_images.append(images[index])
```
mschrimpf, line 248: I'm confused about the difference between `images` and `image_paths`. I thought we could just `output_images.append(image_path)` but I'm not sure how the `images` are different

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -125,31 +177,99 @@ def register_stimulus_set_hook(self, hook):
...
+            else:
+                # translate images according to microsaccades if we are using microsaccades
+                image, image_shape = self.get_image_with_shape(images[index])
+                microsaccade_location_pixels = self.select_microsaccade(image_path=image_path,
+                                                                        trial_number=trial_number,
+                                                                        image_shape=image_shape)
+                return_string = True if isinstance(images[index], str) else False
+                output_images.append(self.translate_image(image, microsaccade_location_pixels, image_shape,
+                                                          return_string))
```
mschrimpf, line 257: kwargs pls

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -161,11 +281,14 @@ def _pad(self, batch_images, batch_size):
     def _unpad(self, layer_activations, num_padding):
         return change_dict(layer_activations, lambda values: values[:-num_padding or None])
 
-    def _package(self, layer_activations, stimuli_paths):
+    def _package(self, layer_activations, stimuli_paths, require_variance):
```
mschrimpf, line 284:
```suggestion
    def _package(self, layer_activations, stimuli_paths, require_variance: bool):
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -182,17 +305,25 @@ def _package(self, layer_activations, stimuli_paths):
             for coord in neuroid_coords:
                 neuroid_coords[coord][1] = np.concatenate((neuroid_coords[coord][1], layer_assembly[coord].values))
             assert layer_assemblies[0].dims == layer_assembly.dims
-            for dim in set(layer_assembly.dims) - {'neuroid'}:
-                for coord in layer_assembly[dim].coords:
-                    assert (layer_assembly[coord].values == nonneuroid_coords[coord][1]).all()
+            for coord, dims, values in walk_coords(layer_assembly):
+                if set(dims) == {'neuroid'}:
+                    continue
+                assert (values == nonneuroid_coords[coord][1]).all()
```
mschrimpf, line 311: these are equivalent right? Just a style rewrite?

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -182,17 +305,25 @@ def _package(self, layer_activations, stimuli_paths):
...
+
         neuroid_coords = {coord: (dims_values[0], dims_values[1])  # re-package as tuple instead of list for xarray
                           for coord, dims_values in neuroid_coords.items()}
         model_assembly = type(layer_assemblies[0])(model_assembly, coords={**nonneuroid_coords, **neuroid_coords},
                                                    dims=layer_assemblies[0].dims)
         return model_assembly
 
-    def _package_layer(self, layer_activations, layer, stimuli_paths):
-        assert layer_activations.shape[0] == len(stimuli_paths)
+    def _package_layer(self, layer_activations, layer, stimuli_paths, require_variance=False):
```
mschrimpf, line 319:
```suggestion
    def _package_layer(self, layer_activations: np.ndarray, layer: str, stimuli_paths: List[str], require_variance: bool = False):
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -209,17 +340,34 @@ def _package_layer(self, layer_activations, layer, stimuli_paths):
...
+                      pixels_or_degrees='pixels',
+                      dim=1)),
+                  'microsaccade_shift_x_degrees': ('presentation', self.unpack_microsaccade_coords(
+                      np.unique(stimuli_paths),
+                      pixels_or_degrees='degrees',
+                      dim=0)),
+                  'microsaccade_shift_y_degrees': ('presentation', self.unpack_microsaccade_coords(
+                      np.unique(stimuli_paths),
+                      pixels_or_degrees='degrees',
+                      dim=1)),
```
mschrimpf, line 359: please add a comment explaining what is going on here. This should probably be extracted into a method `_build_microsaccade_coords`, i.e.
```
coords = {**{'stimulus_path': ('presentation', stimuli_paths),
                 **self._build_microsaccade_coords(...),
                'neuroid_num': ('neuroid', ...
                ...
                } 
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
         wrapper.register_batch_activations_hook = self.register_batch_activations_hook
         wrapper.register_stimulus_set_hook = self.register_stimulus_set_hook
 
+    def select_microsaccade(self, image_path: str, trial_number: int, image_shape: Tuple[int, int]
+                            ) -> Tuple[float, float]:
+        """
+        A function for generating microsaccade locations. The function returns a tuple of pixel shifts expanding from
```
mschrimpf, line 385:
```suggestion
        A function for generating a microsaccade location. The function returns a tuple of pixel shifts expanding from
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+        the center of the image.
+
+        Microsaccade locations are placed within a circle, evenly distributed across the entire area in a spiral,
+        from the center to the circumference. We keep track of microsaccades both on a pixel and visual angle basis,
+        but only pixel values are returned. This is because shifting the image using cv2 requires pixel representation.
+        """
+        # avoid duplicating this computation repeatedly
```
mschrimpf, line 392:
```suggestion
        # if we already computed `self.microsaccades`, we can just index into it
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+        if image_path in self.microsaccades.keys():
+            return self.microsaccades['pixels'][image_path][trial_number]
```
mschrimpf, line 394: this is odd. Why do we index with `trial_number` here, but with `trial_number - 1` in L429?

One way to catch this would be the following structure (this is imo also more intuitive to read):
```
if image_path not in self.microsaccades:  # make sure self.microsaccades is built
    ... # compute self.microsaccades

return self.microsaccades['pixels'][image_path][trial_number]  # or -1, whichever is correct
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+
+        if image_shape[0] != image_shape[1]:
+            self._logger.debug('Warning: input image is not a square. Image dimension 0 is used to calculate the '
```
mschrimpf, line 397: is this a warning or debug information? :) there is `self._logger.warning` for more serious warnings

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+                               'extent of microsaccades.')
+
+        # compute the maximum radius of microsaccade extent in pixel space
+        if self._visual_degrees is None:
+            raise AssertionError('self._visual_degrees is not set by the ModelCommitment, but microsaccades '
+                                 'are in use. Set activations_model visual degrees in your commitment after defining '
+                                 'your activations_model. For example, self.activations_model.set_visual_degrees'
+                                 '(visual_degrees). For detailed information, see '
+                                 ':meth:`~brainscore_vision.model_helpers.activations.ActivationsExtractorHelper.'
+                                 '__call__`,')
```
mschrimpf, line 407:
```suggestion
        assert self._visual_degrees is not None, ('self._visual_degrees is not set by the ModelCommitment, but microsaccades '
                                 'are in use. Set activations_model visual degrees in your commitment after defining '
                                 'your activations_model. For example, self.activations_model.set_visual_degrees'
                                 '(visual_degrees). For detailed information, see '
                                 ':meth:`~brainscore_vision.model_helpers.activations.ActivationsExtractorHelper.'
                                 '__call__`,')
```

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+
+        # to keep consistent with number_of_trials, we count trial_number from 1 instead of from 0
+        self.microsaccades['pixels'][image_path] = selected_microsaccades['pixels']
+        self.microsaccades['degrees'][image_path] = selected_microsaccades['degrees']
+        return selected_microsaccades['pixels'][trial_number - 1]
+
+    def unpack_microsaccade_coords(self, stimuli_paths: np.ndarray, pixels_or_degrees: str,
+                            dim: int):
+        """Unpacks microsaccades from stimuli_paths into a single list to conform with coord requirements."""
+        # Python 3.7 does not support typing.Literal
```
mschrimpf, line 434: ?

mschrimpf: I don't get what this comment bears on

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+        assert pixels_or_degrees == 'pixels' or pixels_or_degrees == 'degrees'
+        unpacked_microsaccades = []
+        for stimulus_path in stimuli_paths:
+            for microsaccade in self.microsaccades[pixels_or_degrees][stimulus_path]:
+                unpacked_microsaccades.append(microsaccade[dim])
+        return unpacked_microsaccades
+
+    def calculate_pixels_per_degree_in_image(self, image_width_pixels: int) -> float:
```
mschrimpf, line 442: related to many methods, but it might be good to group all the micro-saccade stuff into their own class, e.g. `MicrosaccadeHelper` or so

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+        """
+        for temporary_file_path in temporary_file_paths:
+            if isinstance(temporary_file_path, str):  # do not try to remove loaded images
+                try:
+                    os.remove(temporary_file_path)
+                except FileNotFoundError:
+                    pass
+
+    @staticmethod
+    def translate(image, shift: Tuple[float, float], image_shape: Tuple[int, int]):
```
mschrimpf, line 468: `-> ?`

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -231,6 +379,115 @@ def insert_attrs(self, wrapper):
...
+        # translation matrix
+        M = np.float32([[1, 0, shift[0]], [0, 1, shift[1]]])
+
+        # Apply translation, filling new line(s) with line(s) closest to it(them).
+        translated_image = cv2.warpAffine(image, M, (cols, rows), flags=cv2.INTER_LINEAR,  # for sub-pixel shifts
+                                          borderMode=cv2.BORDER_REPLICATE)
+        return translated_image
+
+    @staticmethod
+    def get_image_with_shape(image: Union[str, np.ndarray]):
```
mschrimpf, line 479: `-> ?`

brainscore_vision/model_helpers/activations/core.py
```diff
@@ -261,16 +518,38 @@ def lstrip_local(path):
...
     for column in stimulus_set.columns:
-        assembly[column] = 'stimulus_id', stimulus_set[column].values
-    assembly = assembly.stack(presentation=('stimulus_id',))
+        repeated_values = np.repeat(stimulus_set[column].values, replication_factor)
+        assembly = assembly.assign_coords({column: ('presentation', repeated_values)})  # assign multiple coords at once
+        all_columns.append(column)
+
+    index = all_columns + ['microsaccade_shift_x_pixels', 'microsaccade_shift_y_pixels',
+                           'microsaccade_shift_x_degrees', 'microsaccade_shift_y_degrees']
+    assembly = assembly.set_index(presentation=index)  # assign MultiIndex
```
mschrimpf, line 552: could we not do something like this?
```
    presentation_coords = all_columns + [coord for coord, dims, values in walk_coords(assembly['presentation'])]
    assembly = assembly.set_index(presentation=list(set(presentation_coords)))  # assign MultiIndex
```

brainscore_vision/model_helpers/brain_transformation/neural.py
```diff
@@ -23,21 +23,30 @@ def __init__(self, identifier, activations_model, region_layer_map, visual_degre
     def identifier(self):
         return self._identifier
 
-    def look_at(self, stimuli, number_of_trials=1):
+    def look_at(self, stimuli, number_of_trials=1, require_variance: bool = False):
+        """
+        :param number_of_trials: An integer that determines how many repetitions of the same model performs.
```
mschrimpf, line 28: of the same..? I guess "of the same image the model performs" right?

tests/test_model_helpers/activations/test___init__.py
```diff
@@ -223,6 +285,13 @@ def test_exact_activations(pca_components):
...
     path_to_expected = Path(__file__).parent / f'alexnet-rgb-{pca_components}.nc'
     expected = xr.load_dataarray(path_to_expected)
+
+    # Originally, the `stimulus_path` Index was used to index into xarrays in Brain-Score, but this was changed
+    #  as a part of PR #492 to a MultiIndex to allow metadata to be attached to multiple repetitions of the same
+    #  `stimulus_path`. Old .nc files need to be updated to use the `presentation` index instead of `stimulus_path`,
+    #  and instead of changing the extant activations, this test was simply modified to simulate that.
+    expected = expected.rename({'stimulus_path': 'presentation'})
+
     assert (activations == expected).all()
```
mschrimpf, line 297: I expected a test here that tests the exact values for `microsaccade_shift_x_pixels`, `microsaccade_shift_y_pixels`, `microsaccade_shift_x_degrees`, `microsaccade_shift_y_degrees`. The function `test_require_variance_has_shift_coords` only tests the rather trivial number of values, but not that the values are set correctly

mschrimpf (approved):
Looks good once comments are addressed and tests pass. Especially adding unit tests for the exact values are essential imo.

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: I believe that we explicitly should not do this, right? (similarly to number_of_trials), as if the data are stored, microsaccades will exist in those stored arrays already

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: I believe so (I believe you made this change in person though)

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: Yes, oops. A number of outdated comments in this file.

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: for non-tf models `images` at this point are already tensors. added a comment

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: debug information. fixed

brainscore_vision/model_helpers/activations/core.py
benlonnqvist: While these lines are not related to this PR per se - I see that this suggestion is simpler in the sense that it repeats less arguments. In my opinion, though, it's meaningfully more difficult to read. I've changed to:
```
        if isinstance(stimuli, StimulusSet):
            function_call = functools.partial(self.from_stimulus_set, stimulus_set=stimuli)
        else:
            function_call = functools.partial(self.from_paths, stimuli_paths=stimuli)
        return function_call(
            layers=layers,
            stimuli_identifier=stimuli_identifier,
            require_variance=require_variance)```
which I think is a nice middle ground between readability and  reduced repetition

brainscore_vision/model_helpers/activations/core.py
mschrimpf: the `stored` in the function name just refers to the `@store_xarray` decorator which will still run the function if no result has been stored.

mschrimpf:
All tests on Jenkins (and Travis) are green. Travis reports a failure because it gets killed (likely runtime, unrelated to this PR)

PR #514, opened 2024-01-19
brain-score.org (user:424) | Add new plugin(s): models: [alexnet vit] 

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 50. Plugins added: models: [alexnet vit]


deirdre-k:
Submits the same model/files as #515, very similar but not identical. Assuming the second submission is preferred.

PR #518, opened 2024-01-25
Updated benchmark identifiers to remove lab references

shehadak:
#402 removed lab identifiers from the plugin registries, but it did not remove the lab identifiers defined within their corresponding `Benchmark` instances. This PR updates the `Benchmark.identifier` attributes to match the plugin registries by removing lab identifiers.

This is necessary for scoring purposes, as the model registry identifiers need to match the identifiers in the database, which in turn need to match the `Benchmark.identifier` attribute.


mschrimpf:
We can only make this change if we change the database identifiers at the same time -- all current scores refer to the old `<lab>.<firstauthor>` format, otherwise we break this link.

shehadak:
> We can only make this change if we change the database identifiers at the same time -- all current scores refer to the old `<lab>.<firstauthor>` format, otherwise, we break this link.

You are correct. The database identifiers were updated because they also didn't match the `benchmark_registry` identifiers, so we're not breaking any links.

brainscore_vision/benchmarks/imagenet/benchmark.py
```diff
@@ -22,7 +22,7 @@ def __init__(self):
         self._stimulus_set = stimulus_set
         self._similarity_metric = load_metric('accuracy')
         ceiling = Score(1)
-        super(Imagenet2012, self).__init__(identifier='fei-fei.Deng2009-top1', version=1,
+        super(Imagenet2012, self).__init__(identifier='Deng2009-top1', version=1,
```
mschrimpf, line 25: Not necessary, but maybe you could tackle this in the same PR: we changed the names of engineering/ML benchmarks to the more canonical community name, e.g. `ImageNet-top1` for this case: https://github.com/brain-score/vision/blob/65e2c4608a00bff6cd791f920498d19eab7ac33a/brainscore_vision/benchmarks/imagenet/__init__.py#L4
It would make sense to update the identifier outside the registry too (will require a database update which I can take care of)

brainscore_vision/benchmarks/imagenet_c/benchmark.py
```diff
@@ -82,10 +82,10 @@ def __init__(self, noise_category, sampling_factor=10):
         self.noise_types = self.noise_category_map[noise_category]
 
         ceiling = Score(1)
-        super(Imagenet_C_Category, self).__init__(identifier=f'dietterich.Hendrycks2019-{noise_category}-top1',
+        super(Imagenet_C_Category, self).__init__(identifier=f'Hendrycks2019-{noise_category}-top1',
```
mschrimpf, line 85: in case you'd like to update the ML benchmark identifiers, this is another one of those

brainscore_vision/benchmarks/marques2020/test.py
```diff
@@ -96,11 +96,11 @@ def test_ceilings(benchmark, expected):
...
-               'Marques2020_receptive_field': 'alexnet-dicarlo.Marques2020_receptive_field.nc',
-               'Marques2020_orientation': 'alexnet-dicarlo.Marques2020_orientation.nc',
-               'Marques2020_spatial_frequency': 'alexnet-dicarlo.Marques2020_spatial_frequency.nc',
-               'Marques2020_size': 'alexnet-dicarlo.Marques2020_size.nc',
+        files={'Marques2020_blank': 'alexnet-Marques2020_blank.nc',
+               'Marques2020_receptive_field': 'alexnet-Marques2020_receptive_field.nc',
+               'Marques2020_orientation': 'alexnet-Marques2020_orientation.nc',
+               'Marques2020_spatial_frequency': 'alexnet-Marques2020_spatial_frequency.nc',
+               'Marques2020_size': 'alexnet-Marques2020_size.nc',
                'movshon.FreemanZiemba2013_properties': 'alexnet-movshon.FreemanZiemba2013_properties.nc',
```
mschrimpf, line 104: `movshon.` is still here, but above `dicarlo.` was changed. FWIW I think these could all be kept the same if you don't want to deal with S3 file renaming

brainscore_vision/models/regnet/requirements.txt
```diff
@@ -0,0 +1,2 @@
+torchvision
```
mschrimpf, line 1: wrong PR

brainscore_vision/models/tv_efficientnet-b1/requirements.txt
```diff
@@ -0,0 +1,2 @@
+torch
```
mschrimpf, line 1: wrong PR

brainscore_vision/models/tv_efficientnet-b1/setup.py
```diff
@@ -1,25 +0,0 @@
-#!/usr/bin/env python
```
mschrimpf, line 1: wrong PR

mschrimpf (approved):
looks good, except the model files that should not be in this PR

brainscore_vision/benchmarks/imagenet_c/benchmark.py
shehadak: Updated ML benchmarks in recent commits.

brainscore_vision/models/tv_efficientnet-b1/requirements.txt
shehadak: Reverted in recent commits.

brainscore_vision/models/regnet/requirements.txt
shehadak: Reverted in recent commits.

brainscore_vision/models/tv_efficientnet-b1/setup.py
shehadak: Reverted in recent commits.

brainscore_vision/benchmarks/marques2020/test.py
shehadak: You're correct. I updated them to keep the lab identifiers for test files.

PR #520, opened 2024-01-26
[TEST] fixing database collisions (sqlite in memory database)

samwinebrake:
Now, a new database file will be created in memory for each testing run to avoid database collisions.

Currently testing...

Dependent on [this PR](https://github.com/brain-score/core/pull/77) in core.

Necessary to think about whether we want tests to run in SQLite instead of postgres.. if set on keeping in postgres, then will have to change method of fixing collisions. Perhaps to new schema for every testing run.


samwinebrake:
Not going to delete mock config until this is confirmed to work/confirmed that mock config not needed elsewhere

samwinebrake:
retrying now that openmind space is available

PR #531, opened 2024-01-31
brain-score.org (user:494) | Add new plugin(s): .ipynb_checkpoints: [model_helpers: [utils: [models: [dbp_resnet50_julios] submission: [

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 61. Plugins added: .ipynb_checkpoints: [model_helpers: [utils: [models: [dbp_resnet50_julios] submission: [


mschrimpf:
@kvfairchild shouldn't this have automerged?

shehadak:
> @kvfairchild shouldn't this have automerged?

Good catch! You're correct that this PR should have been automerged, but it seems like (at the time the merge was attempted), the PR had non-plugin differences with the `master` branch, evident by both the PR name (including `.ipynb_checkpoints`) and the [GitHub statuses page for the PR](https://api.github.com/repos/brain-score/vision/statuses/web_submission_61/add_plugins). This should be fixed by re-merging with `master` and re-triggering the checks.

PR #538, opened 2024-02-12
Added 'public' parsing to website submission PRs

shehadak:
This PR updates the hardcoded `public: True` attribute for submitted plugins by parsing it from the submitted PR's name instead, similar to how user information is relayed from brain-score.web.


kvfairchild (approved):
If we're not going to do it in this PR, can you make an issue for renaming this `public` variable (both here and in core) to `leaderboard_public`? Just so it doesn't fall off the radar.

shehadak:
> If we're not going to do it in this PR, can you make an issue for renaming this `public` variable (both here and in core) to `leaderboard_public`? Just so it doesn't fall off the radar.

Agreed. I [created an issue since changing `public` to `leaderboard_public`](https://github.com/brain-score/brain-score.web/issues/240) will require non-trivial changes in the backend and is beyond the scope of this PR.

PR #575, opened 2024-02-25
Support temporal models by changing the TemporalIgnore to Temporal Aligned

YingtianDt:
The change is compatible with image models (whose response does not contain the time_bin dimension).


.gitmodules
```diff
@@ -4,3 +4,9 @@
 [submodule "brainscore_vision/models/temporal_models/VideoMAE"]
 	path = brainscore_vision/models/temporal_models/VideoMAE
 	url = https://github.com/YingtianDt/VideoMAE.git
+[submodule "brainscore_vision/models/temporal_models/mae_st"]
+	path = brainscore_vision/models/temporal_models/mae_st
+	url = https://github.com/YingtianDt/mae_st.git
+[submodule "brainscore_vision/models/temporal_models/VideoMAEv2"]
+	path = brainscore_vision/models/temporal_models/VideoMAEv2
+	url = https://github.com/YingtianDt/VideoMAEv2.git
```
mschrimpf, line 12: this file should not be at the top level, but rather the model plugin (i.e. dependencies are part of the plugin, not the library)

PR #578, opened 2024-02-26
brain-score.org (user:386) | Add new plugin(s): models: [r50_tvpt]  | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 94. Plugins added: models: [r50_tvpt]


deirdre-k:
Deferring to more recent PR #584

PR #582, opened 2024-02-29
[Draft] Removing lab identifiers from stimulus sets

shehadak:
Currently, the file names for most stimulus sets still contain lab identifiers on S3. However, the stimulus set registry recently removed the lab identifiers in the Brain Score 2.0 refactor. This currently leads to an issue where none of the Marques stimulus sets  can be loaded properly during evaluation:
```
AssertionError: No registrations found for dicarlo.Marques2020_orientation
```

Ideally, the solution would be to rename the stimulus set filenames on S3 to match the current identifiers in the `stimulus_set_registry` (e.g. `dicarlo.Marques2020_receptive_field` --> `Marques2020_receptive_field`), but this requires access to S3 and an effort to ensure every single reference to those stimulus sets in the codebase is corrected. This PR provides a hacky fix to unblock certain problems in scoring website submissions.

### DIFF
--- a/brainscore_vision/benchmarks/objectnet/benchmark.py
+++ b/brainscore_vision/benchmarks/objectnet/benchmark.py
@@ -33,7 +33,7 @@ class Objectnet(BenchmarkBase):
         self._similarity_metric = load_metric('accuracy')
         ceiling = Score(1)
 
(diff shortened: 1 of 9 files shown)


deirdre-k:
Completed with #799

PR #581, opened 2024-02-28
email fix for score_plugins workflow

samwinebrake:
Related to closed PR #579 

Broke that PR up into two parts:

1. (THIS PR) a fix for the current score_plugins workflow that finds brain-score emails based on uid and funnels these emails to jenkins for correct email reporting of submission success/failure
2. (FOR A FUTURE PR) failure emails for failing unit testing jenkins jobs

Everything is tested and working, as long as secrets.AWS_ACCESS_KEY_ID and secrets.AWS_SECRET_ACCESS_KEY are set within vision.

(See previous comments from [here](https://github.com/brain-score/vision/pull/579) for input on choices made in this PR.)

PR #592, opened 2024-03-04
Transfer plugin information by artifact for scoring runs

samwinebrake:
Builds on PR #581 

The output of the `process_submission` job wasn't being transferred to the `run_scoring` job due to github thinking secrets were present (See [here](https://github.com/brain-score/vision/actions/runs/8120774897/job/22249505986) for an example of the following: _Warning: Skip output 'PLUGIN_INFO' since it may contain secret._)

Now, `PLUGIN_INFO` is uploaded securely as an artifact after the `process_submission` job and then downloaded as an artifact in the `run_scoring` job, avoiding github's filtering.

### DIFF
--- a/.github/workflows/score_new_plugins.yml
+++ b/.github/workflows/score_new_plugins.yml
@@ -21,7 +21,7 @@ permissions: write-all
 jobs:
 
   process_submission:
-    name: Check if PR makes changes to /models or /benchmarks
+    name: If triggering PR alters /models or /benchmarks, initiates scoring for relevant plugins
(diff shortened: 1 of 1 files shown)


pellegreene (approved):
Looks like Jenkins CI is failing because of multiple AlexNets being registered, which is due to something I did on my end in Submission Pipeline testing(and I'm fixing it now). 

That aside, this PR looks good to me, I'm comfortable merging it.

kvfairchild (approved):

PR #593, opened 2024-03-05
Remove my duplicate submission

pellegreene:
This just undoes the test submissions I made during pipeline evaluation. Submissions added these four files, so I'm removing them. Please let me know if there's a better way for me to do this than just pulling the files out like this.


mike-ferguson (approved):
@samwinebrake As you know more about submission process, is there an easier way to do this? Otherwise, LGTM

PR #597, opened 2024-03-05
brain-score.org (user:510) | Add new plugin(s): models: [dinov2]  | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 107. Plugins added: models: [dinov2]


deirdre-k:
Model is failing tests because of invalid import, may need to specify a compatible version of transformers:

``` bash
>   from transformers import AutoImageProcessor, Dinov2ForImageClassification
E   ImportError: cannot import name 'Dinov2ForImageClassification' from 'transformers'
```

Closing submission.

PR #613, opened 2024-03-11
adding temporal model helpers

YingtianDt:
Main changes:
1. `temporal` folder that contains the temporal model helpers under `brainscore_vision/model_helpers/activations`
2. change `TemporalIgnore` to `TemporalAligned` under `brainscore_vision/model_helpers/brain_transformations`
3. add `temporal_models` under `brainscore_vision/models` that contains example of how to use temporal model helpers and the submission of video models from torchvision.


brainscore_vision/model_helpers/brain_transformation/temporal.py
```diff
@@ -7,7 +7,105 @@
...
+            **coords
+        },
+    )
+    return ret
+
+
+class TemporalAligned(BrainModel):
     """
     Always output the same prediction, regardless of time-bins.
     Duplicates the LayerMappedModel prediction across time.
```
mschrimpf, line 111:
```suggestion
    Deals with the alignment of time-bins.
    
    If the underlying model does not provide a time dimension in its predictions, this always outputs the same prediction for all requested recording time-bins. More specifically, this duplicates the LayerMappedModel prediction across time.
    
    If the underlying model does provide a time dimension in its predictions, align those time points to the requested recording time-bins.
```

brainscore_vision/model_helpers/brain_transformation/temporal.py
```diff
@@ -30,17 +128,23 @@ def visual_degrees(self) -> int:
...
+            self._logger.debug(f'Repeating single assembly across time bins {self._time_bins}')
+            for time_bin in self._time_bins:
+                time_bin = time_bin if not isinstance(time_bin, np.ndarray) else time_bin.tolist()
+                time_bin_start, time_bin_end = time_bin
+                bin_responses = responses.expand_dims('time_bin_start').expand_dims('time_bin_end')
+                bin_responses['time_bin_start'] = [time_bin_start]
+                bin_responses['time_bin_end'] = [time_bin_end]
+                bin_responses = bin_responses.stack(time_bin=['time_bin_start', 'time_bin_end'])
+                time_responses.append(bin_responses)
+            responses = merge_data_arrays(time_responses)
```
mschrimpf, line 143: suggest extracting into its own method to keep the `look_at` method lean (like you did with `assembly_time_align`)

brainscore_vision/model_helpers/activations/temporal/core/__init__.py
```diff
@@ -0,0 +1,227 @@
+import copy
```
mschrimpf, line 1: I realistically won't be able to make sense of all this code. But if you provide unit tests, I can make sense of those

mschrimpf, line 1: let's add a note that this effectively duplicates the existing non-temporal activations extractor with new functionality for temporal models. And that we hope the two can be unified again in the future or that this new wrapper will supersede the previous one

mschrimpf (commented):
general comments:
* add type hints to your code (e.g. `List[int]`, `str` etc)
* add more unit tests to ensure the code is doing what it's supposed to do -- and will continue to do so after inevitable future maintenance changes

PR #616, opened 2024-03-12
add generic benchmark tests

mschrimpf:
resolve https://github.com/brain-score/vision/issues/605

### DIFF
--- a/brainscore_vision/benchmark_helpers/generic_plugin_tests.py
+++ b/brainscore_vision/benchmark_helpers/generic_plugin_tests.py
@@ -1,0 +1,158 @@
+from pathlib import Path
+from typing import Union, List, Tuple
+from brainio.stimuli import StimulusSet
+# the following import is needed to configure pytest
+# noinspection PyUnresolvedReferences
+from brainscore_core.plugin_management.generic_plugin_tests_helper import pytest_generate_tests
+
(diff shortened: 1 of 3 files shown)


mike-ferguson (approved):
Looks good from a cursory glance, but I am not sure how this fits into the testing slimming down recently done. @samwinebrake this might warrant further conversation if this is too computationally expensive to run on every PR

deirdre-k:
@mschrimpf note to review this

mschrimpf:
> Looks good from a cursory glance, but I am not sure how this fits into the testing slimming down recently done. @samwinebrake this might warrant further conversation if this is too computationally expensive to run on every PR

for reference, it might take up to ~15min for a new benchmark submission (see the timings in the first screenshot). It can be a lot faster, but depends on the number of stimuli and complexity of the benchmark. I suggest we proceed with merging because I think these tests will prevent a couple of downstream issues.

mschrimpf:
one problem is that it seems these tests are now run on every single benchmark that is already there. I'm not sure why that is, I thought `pytest` discovery requires the files to be named `test*` and not e.g. `brainscore_vision/benchmark_helpers/generic_plugin_tests.py`.

[Reference](https://doc.pytest.org/en/latest/reference/reference.html#confval-python_files):
> By default, files matching test_*.py and *_test.py will be considered test modules.

Issue #651, opened 2024-03-18
require_variance missing from the signature of look_at


benlonnqvist:
PR #492 introduced the argument `require_variance` to models' `look_at` methods, but did not change the signature in `model_interface`.

PR #652, opened 2024-03-18
add require_variance to model_interface look_at signature

benlonnqvist:
This PR resolves issue #651.

In addition, this PR adds traversal paths for `require_variance` for non-neural task models (Behavior, ProbabilitiesMapping, Temporal).

### DIFF
--- a/brainscore_vision/model_interface.py
+++ b/brainscore_vision/model_interface.py
@@ -196,8 +196,8 @@ class BrainModel:
         """
         raise NotImplementedError()
 
-    def look_at(self, stimuli: Union[StimulusSet, List[str]], number_of_trials=1) \
-            -> Union[BehavioralAssembly, NeuroidAssembly]:
+    def look_at(self, stimuli: Union[StimulusSet, List[str]], number_of_trials: int = 1,
+                require_variance: bool = False) -> Union[BehavioralAssembly, NeuroidAssembly]:
         """
         Digest a set of stimuli and return requested outputs. Which outputs to return is instructed by the
         :meth:`~brainscore_vision.model_interface.BrainMode.start_task` and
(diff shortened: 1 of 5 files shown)


brainscore_vision/model_interface.py
```diff
@@ -207,6 +207,9 @@ def look_at(self, stimuli: Union[StimulusSet, List[str]], number_of_trials=1) \
             or a list of image file paths
         :param number_of_trials: The number of repeated trials of the stimuli that the model should average over.
             E.g. 10 or 35. Non-stochastic models can likely ignore this parameter.
+        :param require_variance: Allows all models access to different activations to the same stimuli, currently
+            implemented using microsaccades. For more information, see the class
+            :cls:`~brainscore_vision.model_helpers.activations.core.MicrosaccadeHelper`.
```
mschrimpf, line 212:
```suggestion
        :param require_variance: Whether to require models to return different activations for the same stimuli or not. This is typically true in biological measurements due to e.g. micro-saccades and noise. Ideally this would always be the same in the models, but for compute reasons this flag allows specifying the need more directly.
```

deirdre-k:
Hi @benlonnqvist, is this PR still in progress? If not, should it be closed? Thanks for the contributions!

benlonnqvist:
Hi @deirdre-k, thanks for the message. This PR is technically still in progress and while PR #917 fixed the issue about being able to call microsaccades, the signature issue in model_interface / elsewhere still seems to be open. 

We could either:
1. wait for me to get to this at some point in the future (I see this as quite a low priority item for me personally)
2. close this and leave the issue #651 open with a note pointing to this PR as a somewhat-approved way of resolving the issue

Thanks!

PR #653, opened 2024-03-18
exclude `test_convolution_meta` from travis

mschrimpf:
This seems to be the test that keeps killing travis builds, e.g. https://app.travis-ci.com/github/brain-score/vision/jobs/619302581.
There are 135 tests that pass (`.`) and the 136th test is `test_convolution_meta`.

[full list of unit tests in `tests/activations/test___init__.py`](https://github.com/brain-score/vision/files/14635070/test_activations.txt)

### DIFF
--- a/tests/test_model_helpers/activations/test___init__.py
+++ b/tests/test_model_helpers/activations/test___init__.py
@@ -1,9 +1,10 @@
 import functools
-import numpy as np
 import os
+from pathlib import Path
+
+import numpy as np
 import pytest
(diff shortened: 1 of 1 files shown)


deirdre-k:
Hi! Looks related to PR #617, not sure if we should prioritize one over the other?

mschrimpf:
ah I didn't see that. Let's see which of the two turns travis green and we'll merge that one (if they both do, my head will explode)

PR #664, opened 2024-03-19
Changing git diff to common ancestor commit

deirdre-k:
Goes with core [PR #92](https://github.com/brain-score/core/pull/92)

Travis tests do not yet account for autosubmission PRs that have fallen out of date with the main/master branch. When the target branch (main) is ahead of the incoming submission branch, the changed files in main are counted as part of "parse_plugin_changes", and consequently "plugin_only" is ruled False which triggers a full test run. This is causing a timeout in the Travis tests.

This PR is a minimal solution that changes the git diff to compare to where the submission branched off of main, thereby only considering the changes directly from the submission. This removes the effect on "plugin_only" rulings and allows plugins to resume testing without running the full repo-wide suite.


samwinebrake (approved):
LGTM

PR #674, opened 2024-03-21
Increase travis timeout for pip install steps

deirdre-k:
Travis jobs are being killed because of slow `torch` install during the `pip install -e .[test]` step. 

By default, Travis kills builds/jobs that do not produce any output for >10 minutes. Adding `travis_wait` to the install step increases the timeout to 20 minutes, which should bypass the torch issue.


mschrimpf (approved):

.travis.yml
```diff
@@ -14,7 +14,7 @@ before_install:
 # download large files
 - pip install awscli
 install:
-- pip install -e ".[test]"
+- travis_wait pip install -e ".[test]"
```
kvfairchild, line 17: Might consider including `--no-cache-dir`, e.g.
```suggestion
- travis_wait pip install --no-cache-dir -e ".[test]"
```
 to see if it actually reduces torch download time instead of just increasing Travis runtime

mike-ferguson (approved):
looks good, - but won't Travis tests still fail on this PR? Looks like this branch might be running into the issue it is addressing for tests

samwinebrake:
retrigger jenkins

mschrimpf:
fyi current travis error:
```
Downloading torch-1.13.1-cp37-cp37m-manylinux1_x86_64.whl (887.5 MB)
Downloading nvidia_cublas_cu11-11.10.3.66-py3-none-manylinux1_x86_64.whl (317.1 MB)
Downloading nvidia_cuda_nvrtc_cu11-11.7.99-2-py3-none-manylinux1_x86_64.whl (21.0 MB)
Downloading nvidia_cuda_runtime_cu11-11.7.99-py3-none-manylinux1_x86_64.whl (849 kB)
Downloading nvidia_cudnn_cu11-8.5.0.96-2-py3-none-manylinux1_x86_64.whl (557.1 MB)

ERROR: Exception:
Traceback (most recent call last):
  File "/home/travis/virtualenv/python3.7.13/lib/python3.7/site-packages/pip/_vendor/urllib3/response.py", line 438, in _error_catcher

    [...]

  File "/opt/python/3.7.13/lib/python3.7/ssl.py", line 929, in read
    return self._sslobj.read(len, buffer)
ConnectionResetError: [Errno 104] Connection reset by peer
```

deirdre-k:
I've moved the torch install into its own step with a `travis_wait` label. Not a super elegant solution but it should get things moving again. I tested out the solution in draft PR #733 where it appeared to work—torch install took 12 minutes but the build was able to run. 

I applied the same change here to keep everything in one thread and test the fix a second time since this error appears to be inconsistent. If it passes I think we should merge.

PR #676, opened 2024-03-22
Tweak odd-one-out task

linus-md:
Use ReLU'ed model activations


mschrimpf (commented):
@linus-md sorry for missing this PR. We have already run several models without ReLU'ing activations -- should we still merge this and re-run?

linus-md:
I would say this is somewhat up for debate. Martin H. suggested this in the spirit that negative activations are noise. As far as I know most models use relu natively in the penultimate layer so it should not make a difference in most cases. Also if a researcher chooses not to use relu this should maybe be respected. What do you say?

mschrimpf:
> if a researcher chooses not to use relu this should maybe be respected. 

that's exactly how I was thinking about it too. Your original implementation is more flexible and leaves the choice to the modeler.

linus-md:
Ok good I'll close this then.

PR #680, opened 2024-03-26
fix `test_endpoints:TestRunScoring` by creating user 2

mschrimpf:
create admin user `2` for new benchmark tests


mschrimpf:
only failing test is `tests/test_submission/test_db.py::test_db` which requires fixing the test db `brainscore-ohio-test` -- merging

PR #685, opened 2024-03-26
fix identifier metadata on Hebart2023

mschrimpf:
this needs to be simultaneously updated in the database


mschrimpf:
travis failing because of the issue identified in https://github.com/brain-score/vision/pull/674

PR #695, opened 2024-03-27

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 171. Plugins added: models: [eSimCLR_Vanilla_2]


samwinebrake:
already added

PR #698, opened 2024-03-27

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 174. Plugins added: models: [eSimCLR_lmda_01_1]


samwinebrake:
already added

PR #700, opened 2024-03-27

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 176. Plugins added: models: [eMMCR_lmda_01_2]


samwinebrake:
already added

PR #701, opened 2024-03-27

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 177. Plugins added: models: [eSimCLR_lmda_01_2]


ThomasYerxa:
Hello, unfortunately the scoring run associated with this PR failed (http://braintree.mit.edu:8080/job/score_plugins/205/parsed_console/), it would be great if that could be re-triggered!

PR #711, opened 2024-03-29
Marques2020 Benchmarks currently fail in production and locally

mike-ferguson:
This PR forces certain `DataArrays` in  `properties_common.py` file to always have the dimensions `(neuroid, presentation`), which the code expects. Previous errors in production and local runs were generated because the dim order was reversed in many cases. 

I have run Alexnet on all Marques2020 benchmarks locally and reproduced the scores on the website. Pixels also works as well. Previously both pixels and Alexnet threw the errors that we saw.

### DIFF
--- a/brainscore_vision/benchmark_helpers/properties_common.py
+++ b/brainscore_vision/benchmark_helpers/properties_common.py
@@ -69,7 +69,7 @@ def get_firing_rates(model_identifier, model, region, stimulus_identifier, numbe
     affine_transformation = firing_rates_affine(model_identifier=model_identifier, model=model, region=region)
     affine_transformation = affine_transformation.values
 
-    activations = record_from_model(model, stimulus_identifier, number_of_trials).transpose()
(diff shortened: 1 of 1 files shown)


mschrimpf:
The [model_interface](https://brain-score.readthedocs.io/en/latest/modules/model_interface.html) does not guarantee the order of dimensions, it only specifies which dimensions have to be there and with which coordinates. So it's possible that we swapped for instance the `neuroid` and `presentation` dimension in the model helper which still satisfies the specification. But for benchmarks like this one that do not ensure the order of dimensions and just take the values, things might have broken.

This means that a `transpose()` call without arguments is also *not viable*! You need to specify which dimensions the DataArray should be transposed to, e.g. `transpose('presentation', 'neuroid')`

samwinebrake:
retrigger jenkins

mike-ferguson:
Merged without Jenkins tests passing, as OM still is not fully back up. Travis tests pass.

PR #768, opened 2024-04-18
brain-score.org submission (user:490) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 228. Plugins added: models: [resnet34]


samwinebrake:
wrong identifier

PR #775, opened 2024-04-19

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 233. Plugins added: models: [eMMCR_Mom_lmda_02_1]


ThomasYerxa:
This PR failed due to a Travis CI timeout, can this be re-triggered?

PR #797, opened 2024-04-25
brain-score.org submission (user:492) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 255. Plugins added: models: [AlexNet_SIN]


samwinebrake:
already added

PR #642, opened 2024-03-14
fix default prefix from image_ to stimulus_

mike-ferguson:
Changes the default value in get_path to be `stimulus_` instead of `image_`

PR #799, opened 2024-04-28
Rename all stimulus sets and S3 files

deirdre-k:
This PR addresses the issue raised in PR #582. This makes all stimulus set identifiers match the keys with which they are registered, and updates the zip and csv paths to point to files in S3 with updated names.

This PR also changes the default prefix for files from "image_"to "stimulus_".


brainscore_vision/data_helpers/s3.py
```diff
@@ -24,7 +24,7 @@ def get_path(identifier: str, file_type: str, bucket: str, version_id: str, sha1
     Finds path of desired file (for .csvs, .zips, and .ncs).
     """
     if filename_prefix is None:
-        filename_prefix = 'image_' if file_type in ('csv', 'zip') else 'assy_'
+        filename_prefix = 'stimulus_' if file_type in ('csv', 'zip') else 'assy_'
```
mike-ferguson, line 27: This is the only change that is in PR #642 -> That PR can be closed since it is addressed here

mike-ferguson (approved):
Looks good (from an initial glance of each of the 101 files changed)

PR #811, opened 2024-05-03
add temporal models (r3m, resnet, dino, mae)

thekej:


YingtianDt (commented):
These look right, but a few things in general:

1. could you rename each folder to have "temporal_model_" as a prefix?
2. now the default inferencer (TemporalInferencer) is used. It will pass the video as a whole into the model, get the features, and evenly distribute the feature from multiple time steps to the whole video duration. If this is what you want, it is fine. There are more choices of inferencers under model_helpers/activations/temporal/core/inferencer.
3. now I think even with the default inferencer you should be able to test these models on any image benchmarks. Could you give it a try to verify if the model is working?

mschrimpf:
overall structure looks good to me. If @YingtianDt thinks this makes sense and once tests pass, fine to merge from my side

brainscore_vision/models/temporal_model_R3M/r3m_model.py
```diff
@@ -0,0 +1,50 @@
...
+        super().__init__()
+        self.r3m = load_r3m(model_name)
+        self.latent_dim = 2048  # resnet50 final fc in_features
+
+    def forward(self, x):
+        return self.get_encoder_feats(x)  # R3M expects image input to be [0-255]
+
+    def get_encoder_feats(self, x):
+        # applies encoder to each image in x: (Bs, T, 3, H, W) or (Bs, 3, H, W)
+        with torch.no_grad():  # TODO: best place to put this?
```
YingtianDt, line 19: The whole `forward` function will be run under `torch.no_grad()`

brainscore_vision/models/temporal_model_mae/model.py
```diff
@@ -0,0 +1,66 @@
...
+import torch as th
+
+from brainscore_vision.model_helpers.activations.temporal.model import PytorchWrapper
+from torchvision import transforms
+
+
+class MAEWrapper(PytorchWrapper):
+    def forward(self, inputs):
+        tensor = th.stack(inputs)
+        tensor = tensor.permute(0, 2, 1, 3, 4)
```
YingtianDt, line 11: One question: the mae (and resnet) seems to be a image model, but here there are videos passing in. How does it work?

brainscore_vision/models/temporal_model_mcvd/setup.py
YingtianDt, line 1: Does this model only work on GPU? 
If not, can you modify the code so that the model use CPU if GPU is not available?

YingtianDt (commented):
Looks good overall. But some image models are "converted" to video models by reshaping the batch&time dimensions. I prefer to add them instead as image models and will provide a way to convert them into video models. 
I will talk with @mschrimpf on this.

YingtianDt:
1. temporal_model_vjepa: can you support model loading to cpu?
2. temporal_model_fitvid/fitvid_sim: error `RuntimeError: The size of tensor a (8) must match the size of tensor b (15) at non-singleton dimension 3` when feeding in a video of 120 frames.
3. temporal_model_dfm_lstm: `AssertionError: Input width (114) doesn't match model (64).` -- can you add a resizing function to the processing function?
4. temporal_model_dino_lstm: '  File "/home/ytang/workspace/brain-score/brainscore_vision/models/temporal_model_dinov2/model.py", line 20
    transforms.CenterCrop(224)
    ^^^^^^^^^^^^^^^^^^^^^^^^^^
SyntaxError: invalid syntax. Perhaps you forgot a comma?'
5. temporal_model_dino_lstm: 'KeyError: 'dinov2''

thekej:
> 1. temporal_model_vjepa: can you support model loading to cpu?
> 2. temporal_model_fitvid/fitvid_sim: error `RuntimeError: The size of tensor a (8) must match the size of tensor b (15) at non-singleton dimension 3` when feeding in a video of 120 frames.
> 3. temporal_model_dfm_lstm: `AssertionError: Input width (114) doesn't match model (64).` -- can you add a resizing function to the processing function?
> 4. temporal_model_dino_lstm: '  File /home/ytang/workspace/brain-score/brainscore_vision/models/temporal_model_dinov2/model.py, line 20
>    transforms.CenterCrop(224)
>    ^^^^^^^^^^^^^^^^^^^^^^^^^^
>    SyntaxError: invalid syntax. Perhaps you forgot a comma?'
> 5. temporal_model_dino_lstm: 'KeyError: dinov2'

1- are you referring to the weights? because we are not pushing the model to cuda in our implementation.
2- can you send me a trace of the error so i can reproduce it. what fps were you using?
3- There is already a resizing function in the code. I am not sure what shape you are providing to the model. Here is an execution log:

Loading stimuli: 100%|█████████████████████████████| 10/10 [00:00<00:00, 35.79it/s]
activations: 100%|███████████████████████████████| 8/8 [00:59<00:00,  7.38s/it]
Packaging layers: 100%|█ ███████████████████████████| 1/1 [00:00<00:00,  5.80it/s]

4- fixed.
5- Please update the transformers package on your local branch.

KartikP:
@thekej Thank you for this very impressive PR! I’m personally looking forward to see how the temporal models perform.  

We’ve noticed that this PR has numerous models, and as a result is triggering very large unit tests. This is causing long testing times, and if one plugin fails, the whole PR will fail.  

Could you try to break this PR into smaller ones? This way, we can test each subset of plugins incrementally and whenever you’re ready, we can merge models as they pass.

For now, we’ve added a `Skip` label to the PR so that frequent commits to a large PR doesn’t trigger new large unit tests.

PR #848, opened 2024-05-20
test_db.py bug fix

samwinebrake:
Adjusts the amount of benchmarks found in the test to the actual number (7 instead of 6).


mike-ferguson:
Forced merged as per @samwinebrake

PR #618, opened 2024-03-12
Add Ferguson2024 benchmark (competition, honorable mention)

mike-ferguson:
This PR is the Ferguson2024 benchmark PR. This benchmark consists of human psychophysics data collected from January-February 2024, dealing specifically with the search asymmetry phenomenon. There are 14 benchmarks, listed here:

`benchmarks = [ 'circle_line', 'color', 'convergence', 'eighth', 'gray_easy', 'gray_hard', 'half', 'juncture', 'lle', 'llh', 'quarter', 'round_f', 'round_v', 'tilted_line']`

TO DO:

Data Plugin Side: -> #894 (merged)
- [x] Packages Stimuli
- [x] Packages Data 
- [x] Tests stimuli and Data
- [x] Create pretraining stimuli
- [x] combine pretraining stimuli into one set
- [x] Update circle_line

Metric Plugin Side -> #882 (merged)

- [x]  Decide on Metric to use
- [x] Create metric.py
- [x] Create metric __init__.py
- [x] Create metric test.py

Benchmark Plugin Side ->  (this PR)
- [x] Implement Linear decoder (trained all all 14 styles)
- [x] add bootstrap code for integral error calculation
- [x] Generate human integral from assembly
- [x] Implement ceiling method
- [x] Precompute all ceilings (compute heavy)
- [x] Finalize `benchmark.py`
- [x] Finalize `__init__.py file`
- [x] Finalize `test.py file`

PR #882, opened 2024-05-29
Added ValueDelta Metric

mike-ferguson:
Adds the new metric, value delta, that is used primarily in the Ferguson2024 benchmarks but can be used for any two scalar values. The metric uses a ceiled sigmoid function to map similarity of two scalars to a number between 1 and 0.

Main Benchmark PR here: #618


brainscore_vision/metrics/value_delta/metric.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
+from brainscore_core import Metric
+from brainscore_vision.metrics import Score
+
+'''
+This metric takes in two floats and gives a score between 0 and 1 based on 
+how far apart the numbers are, using a sigmoid function.
+
```
mschrimpf, line 8:
```suggestion
```

mschrimpf (approved):

PR #899, opened 2024-06-05
fix stimulus set prefix

mschrimpf:
`image_` -> `stimulus_`


mike-ferguson:
This is addressed in #799 and might be dependent on renaming the existing stimuli. I think if this is merged before the stimuli are renamed, the plugin tests will fail for a lot of plugins, but I am not sure - @deirdre-k do we have some sort of circular dependency?

deirdre-k:
Yes, that's exactly right. We have this change in the linked PR, I'll try to push that through today

mschrimpf:
ah I could have sworn I saw this somewhere before. Ok note that this particular change is independent of renaming existing stimuli, but I will close this PR assuming that #799 gets merged very quickly

PR #917, opened 2024-06-14
Add callable microsaccades as intended by #492

benlonnqvist:
PR #492 added microsaccades to brain-score but unintendedly, were only accessible by directly accessing the `activations_model._extractor`. This PR extends the microsaccades to be callable from the `model.look_at()` method (for testing stimuli) and `model.start_task()` method (for fitting stimuli), as intended by the description of PR #492.

In short, this PR simply extends the call signature of `require_variance` and `number_of_trials` a bit further than before.

PR #365 depends on this PR.

- [x] add a test to catch errors with attach_stimulus_set_meta shape when called from from_stimulus_set

PR #925, opened 2024-06-18
Marking test_look_at_neural_V1 memory intense

deirdre-k:
`test_look_at_neural_V1` is running out of memory in Travis, causing tests to fail (see #917) with the following error:

``` bash
FAILED brainscore_vision/model_helpers/generic_plugin_tests.py::test_look_at_neural_V1[resnext101_32x8d_wsl] -
RuntimeError: [enforce fail at alloc_cpu.cpp:75] err == 0. DefaultCPUAllocator: can't allocate memory: 
you tried to allocate 102760448 bytes. Error code 12 (Cannot allocate memory)
```

This PR marks the test `memory_intense`, meaning it will only be run in Jenkins


pellegreene (approved):

mschrimpf:
re-trigger jenkins

PR #937, opened 2024-06-24
Add resnet50_robust_l2_eps1 (rank 48)

SusanWYS:
Converted resnet50_robust_l2_eps1

### DIFF
--- a/brainscore_vision/models/resnet50_robust_l2_eps1/model.py
+++ b/brainscore_vision/models/resnet50_robust_l2_eps1/model.py
@@ -1,0 +1,82 @@
+from brainscore_vision.model_helpers.check_submission import check_models
+from brainscore_vision.model_helpers.activations.pytorch import PytorchWrapper
+from brainscore_vision.model_helpers.activations.pytorch import load_preprocess_images
+import torchvision
+import functools
+import torch
+import os
+from torchvision.models import resnet50
+from brainscore_vision.model_helpers.s3 import load_weight_file
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+"""
+Template module for a base model submission to brain-score
+"""
+
+
+def get_model_list():
+    """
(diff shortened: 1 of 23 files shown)


brainscore_vision/models/resnet50_robust_l2_eps1/model.py
```diff
@@ -0,0 +1,82 @@
...
+from torchvision.models import resnet50
+from brainscore_vision.model_helpers.s3 import load_weight_file
+dir_path = os.path.dirname(os.path.realpath(__file__))
+
+"""
+Template module for a base model submission to brain-score
+"""
+
+
+def get_model_list():
```
mike-ferguson, line 17: You can delete this method, specifically lines 17-24. This was an old method that Martin let us know we do not need anymore in 2.0 converted models.

brainscore_vision/models/resnet50_robust_l2_eps1/test.py
```diff
@@ -0,0 +1 @@
+# Left empty as part of 2023 models migration
```
mike-ferguson, line 1: For all converted models, we want a simple test to make sure the model is added correctly to Brain-Score:
```

import pytest
import brainscore_vision



@pytest.mark.travis_slow
def test_has_identifier():
    model = brainscore_vision.load_model('resnet50_robust_l2_eps1')
    assert model.identifier == 'resnet50_robust_l2_eps1'
```

This will allow(once the model is merged) Brain-Score to test and ensure that the model is located where it should be. You can also delete the comment on line 1.

mike-ferguson (changes requested):
Nicely done! Looks really good - some minor changes requested, but nothing too bad. Congrats on the first conversion :D. We will keep this open and merge it when it reproduces. @deirdre-k I think we should point all of the PRs that we will open to a feature branch, called like `models_25-50` or something like that, and then once they reproduce we can merge them into master all at once.

PR #950, opened 2024-06-25
Ep/add densenet 201 pytorch

pellegreene:


mike-ferguson (approved):

deirdre-k:
Killed in OpenMind during `test_look_at_neural_V1 `, see error:

``` sh
slurmstepd: error: Detected 1 oom-kill event(s) in StepId=37205392.0. Some of your processes may have been killed by the cgroup out-of-memory handler.
srun: error: node107: task 0: Out Of Memory
```

Same error occurred for #996 

We should look into further steps to address this test as it stops a ton of PRs. We've turned it off in Travis, but would be worth looking into improving the efficiency of the test itself. I'll open an issue.

pellegreene:
With the above OOM errors in mind, how do we want to proceed with these PRs? It seems to be happening for at least a handful of models. Does this justify a merge as we've already went through and validated these scores?

pellegreene:
Still seeing a memory error at 256GB. Note to try 512 next.

PR #958, opened 2024-06-26
Add CORnet-S

pellegreene:


brainscore_vision/models/cornet_s/__init__.py
```diff
@@ -0,0 +1,8 @@
+from brainscore_vision import model_registry
+from .helpers.helpers import CORnetCommitment, _build_time_mappings
+from .model import get_model, get_layers, TIME_MAPPINGS
+
+
+model_registry['cornet_s'] = lambda: CORnetCommitment(identifier='cornet_s', activations_model=get_model('cornet_s'),
```
mschrimpf, line 6: shouldn't this be `CORnet-S` for the registry and the identifier? That's how it is displayed on the website

mschrimpf (approved):

brainscore_vision/models/cornet_s/__init__.py
pellegreene: Updated the identifier casing.

deirdre-k:
@pellegreene looks like the issue was due to torch being pinned at a version that isn't available for Python 3.7. I've unpinned it to see if that causes any issues.

PR #972, opened 2024-06-28
conda pin for travis

samwinebrake:
conda updates were causing environments to break. this should now be fixed


samwinebrake:
trying switch to python 3.7.16

PR #993, opened 2024-07-01
Add reset-152_v2

pellegreene:


brainscore_vision/models/resnet-152_v2_pytorch/requirements.txt
```diff
@@ -0,0 +1,2 @@
+torchvision
+torch
```
mike-ferguson, line 2: is `torch` needed in `requirements.txt`?

mike-ferguson (approved):
lgtm, see comment about if we still need `torch` in` requirements.txt`

PR #995, opened 2024-07-01
Add  mobilenet_v2_1-4_244

pellegreene:


brainscore_vision/models/mobilenet_v2_1-4_224_pytorch/requirements.txt
```diff
@@ -0,0 +1,3 @@
+torchvision
+torch
```
mike-ferguson, line 2: Are we leaving in torch in requirements.txt? Not sure what was decided @deirdre-k

mike-ferguson (approved):
lgtm - see comment about torch in `requirements.txt`

brainscore_vision/models/mobilenet_v2_1-4_224_pytorch/requirements.txt
deirdre-k: For now yeah. Once tests are migrated and the conda env approach in ironed out I think we should move it to a project-level dependency, but until then we want to leave it. Thanks for remembering!

mike-ferguson: Sounds good - @pellegreene feel free to ignore my other PR comments regarding this then

PR #683, opened 2024-03-26
fix positional calls to PytorchWrapper in `vit` plugin

mschrimpf:
resolve #682

PR #1016, opened 2024-07-08
Add MobileViT-small

KartikP:
Add the model `MobileViT-small` to model bank


mschrimpf (approved):

mike-ferguson:
@mschrimpf  have you seen this error before in tests:

```
@classmethod
    def _tensor_to_numpy(cls, output):
>       return output.cpu().data.numpy()
E       AttributeError: 'BaseModelOutputWithNoAttention' object has no attribute 'cpu'
```

Seems similar to #683, but not sure what the fix is

mike-ferguson (approved):

mschrimpf:
looks like `output` isn't actually a tensor, but rather the Module producing the tensor. It might be that the layer assignment isn't right; or you'll have to debug to see why no tensor is created.

PR #1027, opened 2024-07-10

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 292. Plugins added: models: [r101_eBarlow_lmda_02_1]


ThomasYerxa:
model scoring failed for this PR (SSL error when trying to access github: http://braintree.mit.edu:8080/job/score_plugins/599/parsed_console/). Can this be re-triggered?

PR #1052, opened 2024-07-12
fix superfluous apostrophe in benchmark identifier

mschrimpf:
this has to be fixed together with the database


brainscore_vision/benchmarks/maniquet2024/test.py
```diff
@@ -9,9 +9,12 @@
 import pytest
 from brainscore_vision import load_benchmark
```
deirdre-k, line 11:
```suggestion
@pytest.mark.private_access
```
Loading benchmark requires S3 credentials

PR #1081, opened 2024-07-21
Added cross_validation_kwargs parameter to the InternalConsistency class

narekvslife:
Added a `cross_validation_kwargs` parameter to the `__init__()` of `InternalConsistency`, which is later passed to `_SplitHalvesConsistency` which was already expecting `cross_validation_kwargs` in it's init.

In my case this was useful because I have a low-repitition setup, where some of the stimuli have been repeated less times than a default number of cross-val splits for internal consistency.

Although internal consistency estimates with less than 5 repititions yield questionable results, this change makes the interface more complete.


mike-ferguson:
@narekvslife Thanks for your PR - it will be reviewed soon!

mschrimpf (approved):

PR #1100, opened 2024-07-30
Add resnet50_v2

pellegreene:


mike-ferguson (approved):
looks good, tests fail with OOM but I think that can be ignored for now

PR #1120, opened 2024-08-01
Fix temporal model mmaction2

deirdre-k:
Adds missing dependency

Still broken: Model uses a lib that depends on decord, only available in py<=3.8
–need to replace or install from source (see [here](https://github.com/salesforce/LAVIS/issues/15))


mschrimpf:
looping in @YingtianDt

PR #1206, opened 2024-08-29
Custom trained models

akgokce:
This PR adds models trained from scratch on ImageNet & ecoset.


mschrimpf (commented):
could you refactor to reduce the number of plugins (i.e. directories in `brainscore_vision/models/`)? Just use one plugin for e.g. all convnexts, one for all resnets etc. Or even just one plugin for everything. You can define multiple models in each plugin.

Also you committed two `.pkl` files by accident, please remove.

mschrimpf (approved):

PR #1208, opened 2024-08-30
Update minor stimulus mismatch in Malania 2007

benlonnqvist:
The stimuli of Malania 2007 had a minor mismatch to the originals which has been fixed here.


benlonnqvist:
@mschrimpf Looks good to me

mschrimpf (approved):

PR #1213, opened 2024-09-04
Coggan behavioral benchmark extension

benlonnqvist:
This PR extends the Coggan behavioral benchmark by adding the labeling task to the Coggan benchmark.

TODOs:

- [x] ImageNet-map Coggan task labels and add to `brain_transformation/behavior`
- [x] Add labeling benchmarks for Coggan
- [x] add engineering benchmark for Coggan
- [x] verify / unify metric use across benchmarks to standard brain-score metric format
- [x] make condition-wise accuracy distance metric (@ddcoggan)
- [x] clean up data handling in `__call__`
- [x] general cleanup of unneeded imports etc.
- [x] optimize accuracydistance over conditions
- [x] add group ceiling
- [x] add chance level adjustment
- [x] check test scores


KartikP:
@benlonnqvist My apologies for closing and reopening your PR. We noticed that the unit test was running out of memory. This should fix it.

benlonnqvist:
Thanks @KartikP, this PR is still in draft status and at least some tests will definitely not pass because the PR is not complete. While I'm not sure why that might cause the OOM issues, it's certainly possible. When the PR is in a state where I think the tests should (barring bugs) pass, I'll mark as ready for review!

KartikP:
That sounds great with me!

As to why your PR is causing OOM issues, it is because changes to `model_helpers` triggers generic tests in a subset of predefined models ([seen here](https://github.com/brain-score/core/blob/ac8919031f659ac3a66051f2d5d054b94aec78d1/brainscore_core/plugin_management/test_plugins.py#L14)). One particular model (resnext101_32x8d_wsl) seems to be the issue. We'll have this dealt with over the next few days.

benlonnqvist:
Oh, right. I remember now, I had that with another PR some time ago as well. Let me know if you want me to slice the `model_helpers` label changes to another PR, but either way is fine either way. Happy to wait a bit as well if a fix is on the horizon.

benlonnqvist:
Hi @KartikP, it seems that the PR is still running into an OOM error. Is there any timeline on when it might be fixed? Thanks again!

PR #1221, opened 2024-09-09
NSD shared Subjects Benchmark

thekej:
1. Dataset description: This is a subset of the NSD dataset containing fMRI brain responses from the 1000 images that were shared between subjects. Each image has been shown to 4 subjects (3 trials each).
2. Metric: PLS regression between model activations and brain region responses.
3. Brain Regions Covered: V1, V2, V3, V4, early, high lateral, high ventral, high parietal.


mike-ferguson:
Hi @thekej, thanks for opening a PR! Would you mind updating the description with more details about your benchmark? Once that is done we can take a look!

thekej:
@mike-ferguson please let me know if anything else is necessary.

kvfairchild:
The model identifier 'mvimgnet_rf' already exists in the database. Please verify and consider renaming with a unique identifier.

akgokce:
Thanks for adding the NSD dataset to BrainScore!

Generally, benchmark data packaging scripts are also included to illustrate how the raw data is formatted. They can also facilitate running local benchmarks via BrainScore once someone gets access to the dataset from the NSD project website.

I was wondering if you were planning to add these scripts or if there are any plans to release them?

Thanks again!

PR #1232, opened 2024-09-12
Put V1 test before behavior test

akgokce:
Tests for Vision Transformer models fail when the line in [result-caching ](https://github.com/brain-score/result_caching/blob/05120fa98a95ed537227bd54be4c04bf6d9af098/result_caching/__init__.py#L336) tries to merge behavioral test results and V1 neural results. The reason is the `neuroid` dimensions do not match: the behavioral array has `channel_x` and `channel_y` dimensions whereas V1 outputs only have `embedding` [see here](https://github.com/brain-score/vision/blob/07c835cf87fd1e312fb07407b66611c157962064/brainscore_vision/model_helpers/activations/core.py#L279). This causes tests in [generic_plugin_tests.py](https://github.com/brain-score/vision/blob/master/brainscore_vision/model_helpers/generic_plugin_tests.py) to fail and the issue disappears when result caching is disabled. Locally, reordering these tests (putting V1 before behavior) resolves the issue.

### DIFF
--- a/brainscore_vision/model_helpers/generic_plugin_tests.py
+++ b/brainscore_vision/model_helpers/generic_plugin_tests.py
@@ -23,24 +23,6 @@ def test_start_task_or_recording(identifier: str):
     can_do = ProbeModel()
     assert can_do.can_start_task(model) or can_do.can_start_recording(model)
 
-
-def test_look_at_behavior_probabilities(identifier: str):
-    model = load_model(identifier)
-    stimuli = fitting_stimuli = _make_stimulus_set()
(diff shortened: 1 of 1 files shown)


mike-ferguson:
@akgokce Thanks for opening this PR, and we will review it ASAP!

mike-ferguson:
also @akgokce just as an FYI, we are right in the middle of migrating our testing suite to AWS, so thanks for being patient with us; our tests above on this PR might not all pass, and that might not have anything to do with your code. I will update you on this PR when our testing is sorted out!

PR #1249, opened 2024-09-17
Fix the version_id changes after moving the weight files; fix dependency issues with temporal models

YingtianDt:


mike-ferguson:
@YingtianDt thanks! Our PR testing is right in the middle of being ported to AWS, so thanks for hanging in there while we sort all of the bugs out.

YingtianDt:
I believe this PR is ready to be merged. @mschrimpf

KartikP:
@YingtianDt Your PR has a couple of issues. Some of them arise from a saturation of disk space for the tests (on AWS execution tests), which we will increase for you. But it appears `temporal_model_mmaction2` is failing your plugin-specific tests (on the legacy unit test plugins).

@mschrimpf All generic tests have failed. Is this because this is just a "different class" of models. Should anything be changed in terms of how `unittest_plugins` is run?

mschrimpf:
> All generic tests have failed. Is this because this is just a "different class" of models. Should anything be changed in terms of how unittest_plugins is run?

The models should still pass on the generic tests. @YingtianDt any ideas why these seem to fail?

PR #1261, opened 2024-09-23
solve ssl error

mike-ferguson:
Should fix SSL error in Braintree


KartikP (approved):
LGTM. Should be benign to add.

KartikP:
@mike-ferguson I am comfortable force merging this. What do you think?

PR #1282, opened 2024-09-30
brain-score.org submission (user:444) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 422. Plugins added: models: [resnet50_11ad3fa6]


mike-ferguson:
@KartikP agreed, I think it is fine

KartikP (approved):

PR #1291, opened 2024-10-01
user_id check only on scoring run

samwinebrake:
Currently a user_id check is running when the `resolve_models_benchmarks` function is called which is unnecessary.


KartikP (approved):
LGTM

PR #1303, opened 2024-10-08
Fix trigger scoring run bug

samwinebrake:
This bug fix addresses 2 separate issues:

1. When long emails are encrypted, `openssl enc -aes-256-cbc -a` generates a Base64-encoded output, which includes newline characters by default to fit within a certain line width. `-A` was added to output the Base64 encoding as a single line.
2. When dealing with artifacts in v4, artifacts must have original names and can't overwrite each other.


KartikP (approved):
LGTM!

PR #1314, opened 2024-10-10
Allow region_layer_map json to be loaded into BrainModel during commitment

KartikP:
This allows a region_layer_map dictionary to be loaded during ModelCommitment.

Region layer maps need to be in a folder in the model's directory called `region_layer_map` and in a file called `${MODEL_IDENTIFIER}.json`.

For example, if you want to add the layer map for `alexnet`.

You would create a `/alexnet/region_layer_map/alexnet.json` file that contains the region_layer_map dictionary.

`
{
"V1": "features.2", 
"V2": "features.7",
"V4": "features.7",
"IT": "features.12"
}
`

Not particularly happy with the function, but i think it gets the job done. Open to feedback.


KartikP:
resnext101_32x8d_wsl_ still being run on `helper` folder changes and causing `OOM`. Need to remove that 😂

brainscore_vision/model_helpers/brain_transformation/__init__.py
```diff
@@ -52,6 +61,26 @@ def __init__(self, identifier,
...
+        try:
+            importer = import_plugin.ImportPlugin(library_root='brainscore_vision', plugin_type='models', identifier=identifier)
+            model_dir = importer.locate_plugin()
+            region_layer_map_path = os.path.join(os.getcwd(), f'vision/models/{model_dir}', f'region_layer_map/{identifier}.json')
+
+            if os.path.exists(region_layer_map_path):
+                with open(region_layer_map_path, 'r') as region_layer_map_file:
+                    return json.load(region_layer_map_file)
+            else:
+                print(f"No region_layer_map file found for {identifier}, proceeding with default layer mapping")
```
deirdre-k, line 78: Should we use logging here?

deirdre-k (changes requested):
Looks good! Just the one comment

brainscore_vision/model_helpers/brain_transformation/__init__.py
deirdre-k: I think it could go either way!

brainscore_vision/model_helpers/brain_transformation/__init__.py
```diff
@@ -52,6 +64,27 @@ def __init__(self, identifier,
...
 
+    def load_region_layer_map_json(self, identifier):
+        '''
+        Attempts to load the region_layer_map from a JSON file in the model's directory
+        If file exists, load JSON. Otherwise, return None and proceed with legacy layer mapping
+        '''
+        try:
+            importer = import_plugin.ImportPlugin(library_root='brainscore_vision', plugin_type='models', identifier=identifier)
+            model_dir = importer.locate_plugin()
+            region_layer_map_path = os.path.join(os.getcwd(), f'vision/models/{model_dir}', f'region_layer_map/{identifier}.json')
```
mschrimpf, line 75: we're using `pathlib`'s `Path` instead of `os.path` throughout the project

mschrimpf, line 75: `os.getcwd()` is dangerous here. What if we're not executing from the brain-score root directory?

brainscore_vision/model_helpers/brain_transformation/__init__.py
```diff
@@ -30,12 +36,18 @@ def __init__(self, identifier,
         self.activations_model._extractor.set_visual_degrees(visual_degrees)  # for microsaccades
         self._visual_degrees = visual_degrees
         # region-layer mapping
+
+        # Attempt to load region_layer_map from JSON, if available
+        region_layer_map = self.load_region_layer_map_json(identifier) if region_layer_map is None else region_layer_map
```
mschrimpf, line 41: I don't like that the `ModelCommitment` has to know about plugins. Can we not load the specification inside the plugin and pass it to the `ModelCommitment` as a parameter?

mschrimpf (changes requested):

brainscore_vision/model_helpers/brain_transformation/__init__.py
KartikP: I'll fix this

KartikP: You're right. I'll test with different roots

brainscore_vision/model_helpers/brain_transformation/__init__.py
KartikP: Modifying the `__init__.py` was the original approach. However, after talking with the others, it seems modifying the init file, especially when there can be multiple model identifiers listed in a single model directory could be messy. It would also make for a messier PR/commit during the model submission.

By adding a check in `ModelCommitment`, all that would be needed would be a single region_layer_map json per model_identifier in the model directory. But I do agree that it disrupts the current flow of the plugin model.

If we do not want to modify the `__init__.py` file of each model (and therefore cannot pass the map to `ModelCommitment` as a parameter), what would you recommend?

mschrimpf: just thinking out loud, but what about an intermediate approach where the `ModelCommitment` expects the full layer mapping, but submissions can specify this to be automatically discovered? Something like
```
model_registry['mymodel'] = lambda: ModelCommitment(
    identifier='mymodel',
    activations_model=get_model(),
    region_layer_map=discover_region_layer_mapping('mymodel'))
```

The `discover_region_layer_mapping` would then first look for a mapping file in the plugin, and if that doesn't exist make one by running the layer selection. This way the plugin `__init__` would not have to be modified but we still maintain the separation between plugins and model helpers.

brainscore_vision/model_helpers/brain_transformation/__init__.py
```diff
@@ -30,12 +36,18 @@ def __init__(self, identifier,
...
+
+        # If region_layer_map is unavailable
         if region_layer_map is None:
             layer_selection = LayerSelection(model_identifier=identifier,
                                              activations_model=activations_model, layers=layers,
                                              visual_degrees=visual_degrees)
             region_layer_map = RegionLayerMap(layer_selection=layer_selection,
                                               region_benchmarks=STANDARD_REGION_BENCHMARKS)
```
KartikP, line 49: Write local JSON file to the model directory. This can be PR'd when they choose to submit.

brainscore_vision/model_helpers/brain_transformation/__init__.py
the--lynx: Martin - Kartik and I just discussed and I would like to propose something a bit more direct that what you suggested.  We create a means to determine if we are running locally (any place but AWS) or in production (AWS). For local runs (again, anything but AWS) if no json file exists, we perform layer mapping and then write the json file. 
 For AWS runs we assume that region_layer_map json file exists and throw an exception if it does not, terminating the scoring run.

This is changing the assumptions of modelCommitment, which hopefully we don't do frequently, but not significantly (and not even as significantly as the solution you propose above).  Additionally, as a benefit, local users, at the time of their PR already have the layer map and have thought about it and checked it against reality.  They can regenerate as many times as desired, simply by removing the json file.

deirdre-k (approved):

brainscore_vision/model_helpers/brain_transformation/__init__.py
mschrimpf: yes that sounds good to me. And keeps the decoupling to a decent extent.

mschrimpf:
Proposal in order to move this forward: you go ahead with this plan, we merge, and I submit another PR that implements what I have in mind.

PR #1361, opened 2024-10-17
Add layer-mapping workflow to trigger Jenkins

KartikP:
This workflow will trigger the `PluginOrchestrator-Layer-Mapping` sequence of Jenkins jobs.

In short, it does the following:
1. Workflow starts on PR into master
2. Triggers are either: web submissions or manual GitHub label (`trigger-mapping`)
3. Status checks using `wait-on-check-action@v1.3.4`
4. When all of the above pass, PR is checked out, and environment is set up
5. Parse changed files to check what model_identifiers have been added and if mapping is needed
6. If model_identifiers are found and mapping is needed, trigger Jenkins to orchestrate layer-mapping

Everything has been tested except the Status Checks because my forked branch does not have the same status checks.

Also needed to create new GitHub secrets for trigger a new job. These secrets have already been added with the prefix `JENKINS_MAPPING_*`

### DIFF
--- a/.github/workflows/map_new_plugins.yml
+++ b/.github/workflows/map_new_plugins.yml
@@ -1,0 +1,144 @@
+name: Perform layer mapping
+
+
+# THIS WORKFLOW ONLY WORKS WHEN NEW_MODEL CHANGES ARE MADE. BENCHMARKS ARE NOT SUPPORTED
+
+on:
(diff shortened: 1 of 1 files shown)


mike-ferguson (approved):
looks good - I am going to defer to Sam's review as he knows more about workflows

PR #1383, opened 2024-10-25
Cap Lonnqvist & Malania score to 1

benlonnqvist:
Lonnqvist and Malania benchmarks didn't cap ceiled the score to 1, and now they do


KartikP (approved):
LGTM

PR #1399, opened 2024-10-28
Layer Mapping Orchestrator (build:114)

kvfairchild:
Layer maps for resnet18_imagenet_full resnet34_imagenet_full resnet50_imagenet_full are being calculated


KartikP (approved):
Successful layer mapping for relevant model identifiers.

PR #1406, opened 2024-10-29
Update imagenet paths

samwinebrake:
The imagenet and imagenet_c benchmarks currently point to braintree stimuli directories. These stimuli have been moved over to s3 and mounted on the docker container that runs scoring.


KartikP (approved):
I assume the 'ImageNet' csv file autogenerated and unlikely to contain manual errors. LGTM me otherwise.

PR #1446, opened 2024-11-12
Automated version bumping and pypi package building

samwinebrake:
This PR adds automated version bumping capabilities to the repo. Now whenever a PR is merged that isn't plugin-only, it will automatically trigger a version-bumping pipeline.

The current versioning system is v._major_._minor_._patch_; merging a PR with no update label will cause the patch version to be updated, adding the `minor update` label will increment the minor version, and adding the `major update` label will increment the major version.

After the version is bumped, 2 different jobs run:

1. Release notes are generated that outline what has changed, and identify the contributor(s) of the changes.
2. The updated package is built and published to PyPI.

_________

**NOTE**: Everything has been tested locally and on a fork, with the exception of the final publishing step to PyPI which will be tested on the merging of this PR.

### DIFF
--- a/.github/workflows/bump_version.yml
+++ b/.github/workflows/bump_version.yml
@@ -15,6 +15,7 @@ jobs:
       checks: write
     outputs:
       NEEDS_BUMP: ${{ steps.check-changes.outputs.needs_bump }}
+      NEW_VERSION: ${{ steps.version-update.outputs.new_version }}
 
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


samwinebrake:
> Also LGTM, quick question: what is the default tag for a non plugin-only PR now? minor? As in any change that is not just a plugin will be a minor, unless otherwise specified to be major via the major tag?

All PRs without a minor or major tag will be just added as a patch!

samwinebrake:
> LGTM, will this workflow need to be implemented in the other repos?

Yes, after this is PRed will implement something similar in core!

PR #1476, opened 2024-11-19
Remove layer map discovery log warning

KartikP:
Previously, we had some logging for when layer mapping could not be loaded. This logging made it seem like there was an actual error rather than a warning/info. See example here: https://github.com/brain-score/vision/issues/533#issuecomment-2457429854

Decided to remove the logging entirely to be consistent with previous version of ModelCommitment. I.e., if no region layer map was provided, it would just silently continue the way it was before.

Additionally, local development (without placing the model in `/models/`) also works but the Error makes it seem like it did not. See below:

image

### DIFF
--- a/brainscore_vision/model_helpers/brain_transformation/__init__.py
+++ b/brainscore_vision/model_helpers/brain_transformation/__init__.py
@@ -79,10 +79,9 @@ class ModelCommitment(BrainModel):
                     self._logger.info(f"Successfully loaded region_layer_map for {identifier}")
                     return json.load(region_layer_map_file)
             else:
-                self._logger.info(f"No region_layer_map file found for {identifier}, proceeding with default layer mapping")
+                self._logger.info(f"No region_layer_map file found for {identifier}. Will proceed with default layer mapping")
                 return None
         except Exception as e:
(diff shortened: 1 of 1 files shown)


mike-ferguson (approved):

deirdre-k (approved):

brainscore_vision/model_helpers/brain_transformation/__init__.py
```diff
@@ -79,10 +79,9 @@ def load_region_layer_map_json(self, identifier):
                     self._logger.info(f"Successfully loaded region_layer_map for {identifier}")
                     return json.load(region_layer_map_file)
             else:
-                self._logger.info(f"No region_layer_map file found for {identifier}, proceeding with default layer mapping")
+                self._logger.info(f"No region_layer_map file found for {identifier}. Will proceed with default layer mapping")
                 return None
         except Exception as e:
-            self._logger.error(f"Error importing model to search for region_layer_map: {e}")
```
mschrimpf, line 85: how about a `debug` statement instead?

mschrimpf (approved):

PR #1482, opened 2024-11-26
add blt vs model

kapoorlab:
Model submission for scoring BLT-VS vision model on brainscore benchmark datasets.


KartikP (approved):
Model LGTM!

PR #1500, opened 2024-11-30
brain-score.org submission (user:407) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 521. Plugins added: models: [vgg16_less_variation_1]


KartikP:
Layer Mapping did not trigger because invalid model submission (specifically the init file is misconfigured)

PR #1593, opened 2024-12-03
brain-score.org submission (user:405) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 631. Plugins added: models: [resnet50_4_all]


ShreyaKapoor18:
@KartikP  if you have the time could you explain to me as to why this fails. I am sorry won't make annoying PR's again

PR #1636, opened 2024-12-11
Updates benchmarks for new S3 bucket/versions

mike-ferguson:
Updates the data plugins (not benchmarks, which do not need to be modified) to allow benchmarks to pull from Brain-Score AWS instead of old DiCarlo AWS.


KartikP (approved):
Requires changes in `data_helpers/s3.py`

KartikP (changes requested):
This should be "brainscore-storage" not with underscore.

KartikP:
Requires https://github.com/brain-score/brainio/pull/55 to work

PR #1645, opened 2024-12-17
fix release notes for version updates

samwinebrake:
The previous tag was not being set correctly when comparing between the previous tag and latest tag to form release notes.


samwinebrake:
(see [here](https://github.com/brain-score/vision/releases/tag/v2.2.3) and [here](https://github.com/brain-score/vision/releases/tag/v2.2.2)... overlap between pr history)

KartikP (approved):
LGTM! Cleaner approach

PR #1649, opened 2024-12-18
Remove new version from latest version list for workflows

samwinebrake:
Previous PR wasn't accounting for the creation of latest version, so was comparing version to itself.


KartikP (approved):
LGTM!

PR #1736, opened 2025-01-20
Submission of a new model called `ReAlnet_cornet`

yilewang:
Based on the `ReAlnet` model, we submitted a new model with the exact `ModelCommitment` module with `CORnet-S`. The reason behind this pull request is to test our `ReAlnet` model with the same condition as the `CORnet-S` model since our model is based on it and used as a benchmark. 

In this submission, the `helpers.py` file is inherited from `CORnet-S` model. We also used the same `time_mappings` from `CORnet-S` model in our `model.py` file.


yilewang:
I modified my `model.py` script and passed the local tests. Now the new model `ReAlnet_cornet` should be able to be tested by brain-scores benchmarks! Thanks!

KartikP (approved):
This looks good to me and consistent with the cornet-s implementation. The main difference appears to be the checkpoints?

Is that correct?

PR #1762, opened 2025-01-30
Add custom mobilenet wrapper

samwinebrake:
Currently, behavioral benchmarks are not working on these mobilenets due to an assertion error related to number of logits results. This wrapper slices the background class out which is not typical in pytorch models.


samwinebrake:
All tests pass, unit plugin test parsing wrong

mike-ferguson (approved):

PR #1817, opened 2025-03-11
Remove yash cornets

mike-ferguson:
As per Yash, remove his models (weights do not work)


KartikP:
Just retriggering to add condition to ignore deleted files in tests.

KartikP (approved):
Good to remove!

PR #1818, opened 2025-03-11
Prevent deletions from triggering scoring

KartikP:
Previously a simple `git diff` was used to identify changed files. Models that are removed, as a result, would be identified as a `CHANGED_FILES`. These would ultimately result in failed builds because the model no longer exists in the updated main branch post-merge.

Added flags to only allow Added (A), Copied (C), Modified (M), Renamed (R) and Changed (T) to trigger scoring. Deletion (D) has not been added. Same change applied in unit test plugins.

### DIFF
--- a/.github/workflows/score_new_plugins.yml
+++ b/.github/workflows/score_new_plugins.yml
@@ -50,10 +50,11 @@ jobs:
           python -m pip install ".[test]"
 
       - name: Save changed files to env var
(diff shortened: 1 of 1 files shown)


mike-ferguson (approved):
Looks good

samwinebrake (approved):
LGTM

PR #1820, opened 2025-03-12
brain-score.org submission (user:660) | (public:False)

kvfairchild:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 744. Plugins added: models: [sewresnet101_1]


KartikP:
This model has already been submitted as `sewresnet101`. Will reach out to submitter about duplicate model.

PR #1852, opened 2025-04-04
Metadata actions

samwinebrake:
This PR creates a new metadata handler workflow that can be called manually, or through the score_new_plugins.yml workflow. It calls upon the handle_metadata.py script present in https://github.com/brain-score/core/pull/138. 

If a PR only contains changes to a metadata.yml file, then only metadata processing will be triggered, and this information will be added to the database. Otherwise, if it is a normal plugin submission, the action will check if metadata already exists in the directory, and if not, it will generate metadata (if not present in the PR already), and PR it in a separate PR, as well as add it to the db. 

NOTE: Actions for generating/validating/adding metadata to db all tested on a fork and dev database.

### DIFF
--- a/brainscore_vision/plugin_management/__init__.py
+++ b/brainscore_vision/plugin_management/__init__.py
@@ -1,0 +1,3 @@
+from .vision_domain_plugin import VisionDomainPlugin
+
+__all__ = ['VisionDomainPlugin'] 
--- a/brainscore_vision/plugin_management/vision_domain_plugin.py
+++ b/brainscore_vision/plugin_management/vision_domain_plugin.py
@@ -1,0 +1,367 @@
+import os
+import re
+import sys
+import logging
+import requests
+import contextlib
+import torch.nn as nn
+from typing import Optional, Dict, Any, List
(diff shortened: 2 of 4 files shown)


mschrimpf (commented):
I like the idea, not enough of a github actions expert to judge the yml files

.github/workflows/score_new_plugins.yml
```diff
@@ -88,20 +166,33 @@ jobs:
...
         uses: actions/upload-artifact@v4
         with:
           name: plugin-info
           path: plugin-info.json
+
+  process_metadata:
+    name: Process metadata changes
+    needs: process_submission
+    if: needs.process_submission.outputs.RUN_METADATA == 'true'
+    uses: ./.github/workflows/metadata_testing.yml
```
KartikP, line 184: Is this supposed to be part of this PR or changed?

KartikP (changes requested):
Change metadata_testing.yml workflow to handle_metadata.yml

KartikP (approved):
Fixed incorrect workflow file name from testing to handle metadata

PR #1855, opened 2025-04-16
adding yolos_tiny

sofiechung:


brainscore_vision/models/scc_yolos_tiny/tempCodeRunnerFile.py
```diff
@@ -0,0 +1 @@
+
```
samwinebrake, line 1: This file can be deleted

brainscore_vision/models/scc_yolos_tiny/test.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+import brainscore_vision
+
+5@pytest.mark.travis_slow
```
samwinebrake, line 4:
```suggestion
@pytest.mark.travis_slow
```

samwinebrake (changes requested):

brainscore_vision/models/scc_yolos_tiny/test.py
samwinebrake: @sofiechung looks like this 5 is causing issues during testing

samwinebrake (approved):

PR #1856, opened 2025-04-16
adding briaai rmbg model

sofiechung:


brainscore_vision/models/briaai_rmbg_1_4/test.py
```diff
@@ -0,0 +1,7 @@
+import pytest
+import brainscore_vision
+
+5@pytest.mark.travis_slow
```
samwinebrake, line 4: Should work if you remove this 5 :)

samwinebrake (approved):

PR #1809, opened 2025-03-05
Model metadata 1-50

mike-ferguson:
Contains Metadata yaml files for the first 50 (sorted alphabetically) models.

HOLD OFF on merging.

PR #1947, opened 2025-06-10
change yml to yaml for certain models

mike-ferguson:
changes YML files to YAML for 4 models that failed metadata actions from PR #1809 (models 1-50)


KartikP:
@mike-ferguson Is CIFAR supposed to be deleted?

PR #2017, opened 2025-10-02
Test brainio lite mini

mike-ferguson:
Changes the Ferguson2024 benchmark to use new Core BrainIO imports


KartikP (approved):
Straight forward and looks good to me!

PR #2162, opened 2025-12-05
alexnet with field of view 4, 12 and 16 visual degrees

KogniJannis:
Introducing alexnet models that have a different field of view than the default 8 visual degrees.


KartikP (approved):
FOV changes lgtm now!

PR #2169, opened 2025-12-08
Corrected model-to-ceiling mapping under explained variance

sandervanbree:
Given that Split-half/Spearman-Brown ceiling is a variance (reliability) ceiling, the following correction was made: Normalize neural scores as r^2 / reliability instead of (r/reliability) ^2.


mschrimpf:
@KartikP we should also go through the neural benchmarks to see if they need to be individually corrected

mschrimpf:
I'm also attaching the email chain for reference.
[email-thread.txt](https://github.com/user-attachments/files/24052049/email-thread.txt)

PR #2179, opened 2025-12-11
brain-score.org submission (user:726) | (public:False)

KartikP:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 1009. Plugins added: models: [convnext_tiny_imagenet1k_HAND_crop]


KartikP:
Hi @mwspratling , I will take a look at this today and report back.

PR #2185, opened 2025-12-15
brain-score.org submission (user:726) | (public:False)

KartikP:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 1013. Plugins added: models: [convnext_tiny_imagenet1k_torchvisionV1]


KartikP:

PR #2364, opened 2026-04-10
Implement dual ridge regression for memory optimization

KartikP:
Ridge and RidgeCV regression now use the dual kernel form when `n_samples < n_features`. This is common in Brain-Score which there are only hundreds of stimuli but thousands to hundreds of thousands of model features. This avoids materializing the coefficient matrix (`n_features`, `n_targets`) which can produce OOM for large models on fMRI benchmarks.

This PR adds `DualRidgeRegression` and `DualRidgeCVRegression` in the `metrics/regression_correlation/metric.py`.

Instead of solving for the full weight matrix, we precompute a small project matrix and apply it to the neural target in chunks.

This PR effectively reduces a coefficient matrix for ViT-L at 148k features on Allen2022 at 52k neuroids stored at 8 bytes from 62GB down to ~1MB.

### Validation
- All 4 Allen2022 surface ridge benchmarks produce identical raw scores
- Identical predictions across alpha values
- RidgeCV selects the sample alpha as sklearn

### DIFF
--- a/brainscore_vision/benchmark_helpers/neural_common.py
+++ b/brainscore_vision/benchmark_helpers/neural_common.py
@@ -31,6 +31,18 @@ class NeuralBenchmark(BenchmarkBase):
         source_assembly = candidate.look_at(stimulus_set, number_of_trials=self._number_of_trials)
         if 'time_bin' in source_assembly.dims and source_assembly.sizes['time_bin'] == 1:
             source_assembly = source_assembly.squeeze('time_bin')  # static case for these benchmarks
+        # Free model weights — activations are extracted, model not used again.
+        # Can't `del candidate` because _run_score holds a reference via `model`.
+        # Instead clear PyTorch parameters directly.
+        import gc
+        try:
+            torch_model = candidate.activations_model._model
+            for param in torch_model.parameters():
+                param.data = param.data.new_empty(0)
+            del torch_model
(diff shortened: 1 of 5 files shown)


KartikP:
Superseded by #2383, which landed the minimal version of this work on 2026-05-26 — master now has `DualRidgeRegression`, `DualRidgeCVRegression` and the `dual_ridge_split` / `dual_ridgecv_split` metrics. That is why this branch shows a conflict: it is stale against its own merged descendant.

The remaining work is not reviving this PR but applying the switch to the benchmarks #2383 did not cover. Evidence that it matters, from `brainscore_resource_usage`:

| benchmark | OOM rate | path |
| --- | --- | --- |
| `Papale2025.V1-ridgecv` | 68% | primal |
| `Papale2025.IT-ridgecv` | 64% | primal |
| `Papale2025.V4-ridgecv` | 62% | primal |
| `Gifford2022.IT-ridgecv` | 60% | primal |
| `Hebart2023_fmri.*-ridgecv` | 32–43% | primal |
| `Zerbe2026_fmri.*-ridgecv` | 23–48% | primal |
| **`Li2026.V1-ridgecv`** | **7%** | **dual** |

Every primal benchmark OOMs at 23–68%; the one dual benchmark sits at 7% despite a 121.9 GB peak. Follow-up PR switching the remainder is in progress.

PR #2366, opened 2026-04-16
Preflight memory

mike-ferguson:
## Summary

Adds a pre-flight memory estimation system that catches OOM errors before committing to a multi-hour benchmark run.

### Problem
Scoring large models on neural benchmarks can take 6+ hours. If the job runs out of RAM, you lose all that time with nothing to show for it. There was no way to know upfront whether a model would OOM on a given benchmark.

### Solution: Metric-aware memory formula

Before scoring begins, the system estimates total RAM needed using a formula matched to each benchmark's regression type:

| Benchmark type | Formula | Notes |
|---|---|---|
| **RDM/RSA** | `activation_gb × 3` | Pairwise distance computation passes through the full activation matrix — overhead scales with features, not stimulus count. Validated: ≤4% error across alexnet/resnet50/ViT. |
| **Ridge/RidgeCV** (`n_features ≤ n_stimuli`, calibrated) | `activation_gb + fixed_benchmark_cost_gb` | sklearn primal solver — gram matrix is `n_stimuli×n_stimuli`, model-independent. Accurate when features are smaller than stimulus count (e.g. alexnet on most benchmarks). Validated: ≤1% error. |
| **Ridge/RidgeCV** (`n_features > n_stimuli`) | `activation_gb × 6` | sklearn switches to SVD of X — overhead ≈ 5× activation, so calibrated fixed cost (measured on small models) severely underestimates. Falls back to ×6 so pre-flight raises `MemoryError` cleanly before the OS kills the container. Validated: −1.5% error for resnet50 and ViT on Gifford2022. |
| **Ridge/RidgeCV** (`n_features ≤ n_stimuli`, no calibration entry) | `activation_gb + n_stimuli²×4B` | Formula fallback using gram matrix size. |
| **PLS** (`-pls`, `-reverse_pls`) | `activation_gb × 7 + fixed_benchmark_cost_gb` | PLS cross-covariance matrices scale with `num_features` — overhead is not model-independent. Multiplier calibrated on 3-model × 2-benchmark grid; worst miss: −12.7% (within 15%). Warning printed at runtime. |
| **Fallback** | `activation_gb × 6` | Used when benchmark type is unrecognised. |

- **`activation_gb`** — model-dependent. Measured by running a single forward pass (the "probe") on 1 stimulus and extrapolating: `num_stimuli × num_features × num_timebins × 4 bytes`
- **`fixed_benchmark_cost_gb`** — benchmark-dependent, model-independent. Stored in `benchmark_costs.json`.

---

### New file: `brainscore_vision/benchmark_helpers/benchmark_costs.json`
Calibration table with fixed overhead costs for 49 neural benchmarks. Used for ridge/ridgecv and PLS benchmarks. RDM benchmarks do not require calibration entries.

### `brainscore_vision/benchmark_helpers/memory.py`
- **`preallocate_memory(model, benchmark)`** — probes the model with 1 stimulus, dispatches to the correct formula based on benchmark type, and raises `MemoryError` if estimate exceeds available RAM
- **Metric-type detection** — three new helpers: `_is_pls_minimum`, `_is_rdm_benchmark`, `_is_ridge_benchmark`. Detection is by identifier suffix (`-pls`, `-reverse_pls`, `-temporal-pls`, `-rdm`, `-ridge`, `-ridgecv`) and/or `isinstance` check for `RSABenchmark`
- **`RSABenchmark` support** — previously raised `TypeError` for RSA/RDM benchmarks; now fully handled with the model-independent `n_stimuli²` formula
- **`MemoryEstimate` dataclass** — added `formula_type` field (`'pls'`, `'rdm'`, `'ridge_formula'`, `'calibrated'`, `'fallback'`) and `rdm_overhead_gb` field; `__str__` renders the correct formula per type
- **PLS overhead multiplier** — reduced from ×10 to ×7; a warning is now printed when PLS formula is used: `WARNING: PLS overhead multiplier (×7) is approximate. Actual usage can vary significantly depending on model feature count and convergence.`
- **`load_calibration()` / `save_calibration()`** — unchanged; automatically loads bundled `benchmark_costs.json`; falls back to `~/.brainscore/benchmark_costs.json`

### `brainscore_vision/__init__.py`
- Pre-flight runs automatically on every `score()` call via `score_benchmark`
- Improved `AssertionError` message for stale activations cache, with the exact `rm` command to fix it

### `scripts/preflight_check.py` *(new)*
Integration test script for a single (model, benchmark) pair:

### `scripts/mem_profile_suite.py`
- Added `--calibrate` mode: runs alexnet on all neural benchmarks to produce the fixed-cost table
- Saves incrementally after each benchmark (crash-safe)
- Added `--resume-from BENCHMARK_ID` to pick up after a crash

### `scripts/validation.py` *(new)*
Runs a 3-model × 3-benchmark grid to validate how accurately the pre-flight estimator predicts actual peak RSS. Reports over/under estimates per pair and writes results to `validation_results.jsonl`.

---

## How to extend the calibration table

```bash
# Run alexnet on all benchmarks and save results
python scripts/mem_profile_suite.py --calibrate --csv ~/calibration.csv

# Resume after a crash
python scripts/mem_profile_suite.py --calibrate --csv ~/calibration.csv \
    --resume-from MajajHong2015public.IT-pls

## Tests (`tests/test_plugin_management/test_memory_precheck.py`)

- **`TestCalibrationIO`** — load/save roundtrip, missing file, corrupt JSON, directory creation
- **`TestCalibratedFormula`** — calibrated, RDM formula, ridge formula fallback, PLS ×7, fallback ×6; OOM detection for each path
- **`TestMemoryEstimateStr`** — `__str__` renders correct formula label per type; `OK` / `OOM LIKELY` status
- **`TestCalibratedIntegration`** — full roundtrip with mock model and benchmark; confirms `score_benchmark` calls `preallocate_memory` before scoring

---

## Still ToDo

- [x] Complete remaining benchmark estimates via EC2 runs (in progress) to populate JSON
```


brainscore_vision/benchmark_helpers/neural_common.py
KartikP, line 257: Should RSABenchmark have a `preallocate_memory` override like NeuralBenchmark and TrainTestNeuralBenchmark?

brainscore_vision/benchmark_helpers/memory.py
```diff
@@ -0,0 +1,496 @@
...
+        train_ss = benchmark.train_assembly.stimulus_set
+        test_ss = benchmark.test_assembly.stimulus_set
+        stimulus_set = train_ss
+        num_stimuli = int(train_ss['stimulus_id'].nunique()) + int(test_ss['stimulus_id'].nunique())
+        num_trials = benchmark._number_of_trials
+        timebins = benchmark.timebins
+        region = benchmark.region
+        visual_degrees = benchmark._visual_degrees
+
+    elif isinstance(benchmark, RSABenchmark):
```
KartikP, line 295: See comment in `neural_common.py` about missing `preallocate_memory` override in RSABenchmark. Seems like if `score_benchmark` flow is run, RSA code path in `memory.py` doesn't override?

tests/test_plugin_management/test_memory_precheck.py
```diff
@@ -0,0 +1,559 @@
...
+        save_calibration({'MajajHong2015.IT-pls': 2.8336}, self._cal_path)
+
+        with patch('psutil.virtual_memory') as mock_vm:
+            mock_vm.return_value.available = 64 * (1024 ** 3)
+            with patch('brainscore_vision.benchmark_helpers.memory._DEFAULT_CALIBRATION_PATH',
+                       self._cal_path):
+                est = preallocate_memory(model, bm, raise_if_oom=False)
+
+        self.assertAlmostEqual(est.fixed_benchmark_cost_gb, 2.8336, places=4)
+        self.assertAlmostEqual(est.total_estimated_gb, est.activation_gb + 2.8336, places=4)
```
KartikP: I think there's a mismatch in the formula here? Should be x7 for PLS factor?

brainscore_vision/__init__.py
```diff
@@ -74,7 +76,21 @@ def _run_score(model_identifier: str, benchmark_identifier: str) -> Score:
...
+            '.ActivationsExtractorHelper._from_paths_stored'
+        )
+        raise AssertionError(
+            f"Activations cache has stale paths — the cached stimulus paths no longer match "
+            f"the current stimulus locations (e.g. temp directory changed between runs).\n\n"
+            f"Fix: delete the stale cache entry and re-run:\n"
+            f"  rm {cache_dir}/identifier={model_identifier},stimuli_identifier=*.pkl\n\n"
+            f"Or to clear the entire activations cache:\n"
+            f"  rm {cache_dir}/*.pkl"
+        )
```
KartikP: Could this overwrite actual downstream AssertionError as the above described "State cache error"? If so, either match specific error pattern rather than any AssertionError or detect stale cache more upstream.

brainscore_vision/benchmark_helpers/memory.py
```diff
@@ -0,0 +1,496 @@
...
+              f"  =  {estimate.total_estimated_gb:.3f} GB total", flush=True)
+    elif formula_type == 'ridge_formula':
+        print(f"  +  {estimate.rdm_overhead_gb:.3f} GB gram matrix ({num_stimuli:,}²×4B)  "
+              f"[no calibration entry — formula estimate]"
+              f"  =  {estimate.total_estimated_gb:.3f} GB total", flush=True)
+    elif formula_type == 'calibrated':
+        print(f"  +  {estimate.fixed_benchmark_cost_gb:.3f} GB benchmark overhead (calibrated)"
+              f"  =  {estimate.total_estimated_gb:.3f} GB total", flush=True)
+    else:
+        print(f"  ×{_OVERHEAD_FACTOR}  =  {estimate.total_estimated_gb:.3f} GB total", flush=True)
```
KartikP: Print to stdout and logs.

KartikP (changes requested):
#### Additional change required in an unmodified file
`brainscore_vision.benchmarks.Benchmark` is a separate ABC from `brainscore_core.benchmarks.Benchmark`, so the new no-op from `Core` doesn't propagate. Most if not all behavioral/engineering benchmarks will likely crash with an `AttributeError` at `_run_score`'s `score_benchmark(benchmark, model)` call.

To resolve this, you probably just have to add the same `def preallocate_memory(self, candidate): pass` to `brainscore_vision/benchmarks/__init__.py` `Benchmark`.


### Actual review
Outside of that, I think the rest of the comments will help resolve this PR. Depending on the follow up from integrating with gated scoring, another recommendation might be to structure the failure signal for easier parsing rather than String parsing. Edit: Following up to this, how does an OOM from a behavior benchmark trigger retry policy in gated scoring. Since behavior and engineering benchmarks do not run through this path, if they run OOM on the smallest EC2 instances (As per gated scoring), will they be just incremented up one level to the next queue?

Additionally, this PR adds a lot of lines changed, but largely for the core Brain-Score developer team to maintain (likely unseen be regular users). If we continue to use this approach, we should consider logging every run in a table so we can refine the `benchmark_costs.json` over time.

Moving forward, we should aim to resolve common reasons for OOM e.g., switching to dual ridge when `n_features > n_stimuli` or incremental/chunked solvers for when feature count is large.

This PR adds an important step towards improving brain-score scoring success and appropriately and cost-effectively allocating compute resources.

KartikP (approved):
LGTM now! Need to come back and recalibrate costs for LAION-fMRI

PR #2380, opened 2026-05-02
brain-score.org submission (user:799) | (public:True)

KartikP:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 1198. Plugins added: models: [deepeisnn_ei_cifar10_vgg19]


KartikP:

KartikP:

PR #2399, opened 2026-06-07
benchmark_helpers: subclass BenchmarkBase + add update_score round-trip tests

KartikP:
The structural followup we should have shipped in the first place. Closes the whack-a-mole sequence (PRs #2396, #2397, #2398 each fixed one scoring-DB schema mismatch as it surfaced at production runtime) with two changes that together prevent future ones.

## 1. Subclass \`BenchmarkBase\`

\`MultiSubjectNeuralBenchmark\`, \`KFoldNeuralBenchmark\`, and \`_MultiSubjectRSABenchmark\` now subclass \`brainscore_vision.benchmarks.BenchmarkBase\` instead of hand-maintaining \`self.identifier = ...\`, \`self.version = ...\`, \`self.bibtex = ...\`, \`self.parent = ...\` separately. The standard property contract is the source of truth.

The \`@property ceiling\` becomes \`def _compute_ceiling(self)\` passed as \`ceiling_func=lambda: self._compute_ceiling()\` — \`BenchmarkBase\` wraps it with \`@store()\` caching, so we also get cross-call ceiling memoization for free (a small bonus on top).

## 2. \`test_db_contract.py\` round-trip tests

New file with 5 tests exercising the exact code paths \`brainscore_core.submission.database.update_score\` takes:

- \`_retrieve_score_center(score)\`  → \`score.item()\` (ceiled column)
- \`_retrieve_score_center(score.raw)\` → \`score.raw.item()\` (raw column)
- \`_retrieve_score_error(score)\` (error column)
- \`wrapper.identifier\`, \`.version\`, \`.parent\`, \`.bibtex\` (DB join keys)

Each test instantiates a wrapper with a tiny \`_SyntheticChild\` mock factory and verifies the produced Score satisfies the contract.

**Every issue we hit after the LAION-fMRI merge would have failed these tests at unit-test time:**

| Issue | Caught by |
|---|---|
| #2396: missing \`version\` attribute | \`assert wrapper.version is not None\` |
| #2396: multi-entry \`BIBTEX\` | \`parse_bib(wrapper.bibtex)\` would fail → caught at module import via #2396 path |
| #2397: \`raw\` set to per-subject array | \`_retrieve_score_center(score.raw)\` raises ValueError |
| #2398: \`raw\` set to Python \`float\` | \`_retrieve_score_center(score.raw)\` raises AttributeError |

## Validation

- 26 existing tests (\`TestRegistry\` + \`TestNonHeadlineFactoriesConstruct\`) still pass
- 5 new tests in \`test_db_contract.py\` pass
- AWS Batch job \`fc86f46e-94f7-4024-8c2c-623da44488c3\` ran alexnet × Zerbe2026_fmri.V1-ood-ridgecv on this branch end-to-end through scoring → \`update_score\` → DB INSERT. **Row 179109 in \`brainscore_score\`**: raw=0.492, ceiled=0.299, error=0.027 (matches our v10 sweep baseline).

## Diff stats

- \`benchmark_helpers/multi_subject.py\`: +29 / −24 (KFold + MultiSubject subclass swap)
- \`benchmarks/laion_fmri/benchmark.py\`: +14 / −11 (RSA wrapper subclass swap)
- \`benchmark_helpers/test_db_contract.py\`: +186 / 0 (new file, 5 tests)

Net structural change is small; the test file is the actual safety net.

### DIFF
--- a/brainscore_vision/benchmark_helpers/multi_subject.py
+++ b/brainscore_vision/benchmark_helpers/multi_subject.py
@@ -33,6 +33,7 @@ import numpy as np
 import xarray as xr
 
 from brainscore_core import Score
+from brainscore_vision.benchmarks import BenchmarkBase
 from brainscore_vision.benchmark_helpers.neural_common import TrainTestNeuralBenchmark
 from brainscore_vision.metric_helpers.bootstrap_error import (
     attach_error,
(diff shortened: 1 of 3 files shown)


KartikP:
Folding into #2398 per request — combining hotfix + structural refactor + tests into one PR.

PR #2413, opened 2026-06-16
brain-score.org submission (user:823) | (public:True)

KartikP:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 1213. Plugins added: models: [convnext_large_clip_laion2b_finelayers]


KartikP:

KartikP:

PR #2452, opened 2026-06-24
brain-score.org submission (user:823) | (public:True)

KartikP:
Plugin-only changes from brain-score.org submission. Generated via Jenkins job create_github_pr, build 1247. Plugins added: models: [vit_large_clip336_openai]


KartikP:

KartikP:

PR #2469, opened 2026-07-30
Add Qwen3.6-27B vision encoder

KartikP:
Adds the Qwen3.6-27B pretrained vision encoder as a Brain-Score Vision model using the existing `PytorchWrapper` and `ModelCommitment`.

- Lazily loads only visual checkpoint tensors
- Exposes six visual transformer blocks
- Adds metadata and focused preprocessing, activation, and loading tests

Tests: `pytest brainscore_vision/models/qwen3_6_27b/test.py -q` (9 passed)


KartikP:

KartikP:

PR #2485, opened 2026-08-11
feat(benchmarks): Zerbe2026 linear-CKA variants with inter-subject ceiling

KartikP:
> **Stacked on #2484.** Base is `kp/ceiling-error`, not `master`. **Merge #2484 first**, then retarget this to `master`. See the merge-order note at the bottom — I lost a PR's contents to exactly this pattern last week.

Draft — the mechanism works end to end on real data, but two decisions are open.

Requested by Yamins (SAB 2026-07-29): does linear CKA predict behavioural metrics better than ridge? Zerbe already carries ridge and RSA variants on the same stimuli, so CKA makes that comparison within-dataset.

## Scope is smaller than it sounds

The metric already existed — `metrics/cka`, registered as `cka` / `cka_cv` / `cka_biased`. Nothing new there. No new benchmark class either: this reuses `RSABenchmark` with the RDM stage bypassed, because CKA compares representations directly and never forms an RDM.

```python
rdm=_passthrough,              # RSABenchmark computes similarity(rdm(model), rdm(neural))
similarity=load_metric("cka"), # CKA needs the raw representations, not RDMs
```

**No cache changes.** Activation reuse is handled upstream by `@store_xarray` and is invisible here — these produce identical scores with the cache on or off.

## Ceiling: pairwise inter-subject CKA

Not the ncsnr proxy the RSA variants use, for two reasons. That ceiler is documented as a bound on how well a model *RDM* can correlate — not the quantity CKA computes — and its inputs are the dataset's published per-voxel reliability rather than anything measured under this metric.

Nili's "each subject vs the mean across subjects" has no CKA analogue either: representations cannot be averaged across subjects because their voxels have no correspondence. The direct analogue is `mean_{i<j} CKA(subject_i, subject_j)`, which puts score and ceiling on the same scale by construction.

## First real score — `alexnet`, `Zerbe2026_fmri.IT-cka`

```
ceiled         0.5232639563817907
raw            0.2504352496354321
ceiling        0.4786021406234721   (10 pairs, 0.4216 – 0.5222)
ceiling_error  0.0112129413120552
```

Scored in the same run, unchanged: `Zerbe2026_fmri.IT-rdm-pearson` = **0.3745153062392548**, bit-identical to its previously recorded value. So the `_passthrough` / `_neutral_ceiler` changes are confined to the CKA path.

## The result worth discussing before merge

```
              raw      ceiling   ceiled
CKA          0.2504    0.4786    0.5233
RSA (rdm)    0.2815    0.7518    0.3745
```

Inter-subject agreement is much weaker under CKA than under RDM correlation (0.479 vs 0.752) on identical data. So alexnet's **raw** CKA is *lower* than its raw RSA, while its **ceiled** CKA comes out 40% *higher* — purely because the denominator is smaller. A metric can look better by having a noisier ceiling. That is exactly the comparability question being asked, and it should be a deliberate decision rather than a side effect.

## Open decisions

1. **Registered surface goes 20 -> 24.** The count guard and module docstring move with it. That is a leaderboard change.
2. **Whether CKA should be ceiled at all**, given the above.

## DB contract

Checked against `brainscore_core.submission.database`, since `BenchmarkInstance` is written `if created` only and a mistake is permanent until a version bump:

- `ceiling.item()` scalar -> `BenchmarkInstance.ceiling`
- `ceiling.attrs['error']` set explicitly -> `ceiling_error` (this is what #2484 fixes for the existing benchmarks)
- `raw` scalar -> `score_raw`; `comment` untouched (`CharField(max_length=1_000)` — the variant metadata would fail the write, not truncate; measured `comment_len=89`)

Extra metadata rides along for later re-normalisation without re-scoring: `ceiling_method`, `ceiling_variants`, `ceiling_pairs` (all 10 labelled), `ceiling_loo_per_subject`. Note `loo_per_subject_mean` equals the pairwise mean by construction for equal subject counts — the per-subject vector is the informative part. `split_half_within_subject` is deliberately `None`: it needs the unaveraged repetition assembly, and the scoring assemblies use `average_repetitions=True`.

`peak_aggregation` 1.0 -> 2.0, since the ceiling holds two assemblies at once where scoring holds one.

**25 non-private tests pass.**

## Merge order

#2484 first, then this. If this merges into `kp/ceiling-error` *after* that branch has already gone into `master`, the changes land on a dead branch and never reach `master`. That is precisely how brain-score/result_caching#23 lost its contents last week — merged into a stacked branch nine seconds after the base merged into master.

### DIFF
--- a/brainscore_vision/benchmark_helpers/multi_subject.py
+++ b/brainscore_vision/benchmark_helpers/multi_subject.py
@@ -380,6 +380,8 @@ class MultiSubjectNeuralBenchmark(BenchmarkBase):
         score.attrs["sem"] = (
             float(values.std(ddof=1) / np.sqrt(len(values))) if len(values) > 1 else 0.0
         )
+        # the recorder reads attrs['error'], not 'sem'; absent means a NULL column
+        score.attrs["error"] = score.attrs["sem"]
         return score
 
     def __call__(self, candidate) -> Score:
@@ -415,6 +417,8 @@ class MultiSubjectNeuralBenchmark(BenchmarkBase):
         ceiling.attrs["sem"] = (
             float(ceil_values.std(ddof=1) / np.sqrt(len(ceil_values))) if len(ceil_values) > 1 else 0.0
         )
+        # the recorder reads attrs['error'], not 'sem'; absent means a NULL column
+        ceiling.attrs["error"] = ceiling.attrs["sem"]
         score.attrs["ceiling"] = ceiling
         for sub_id, s in zip(self._subjects, per_subject_scores):
             score.attrs[sub_id] = s
--- a/brainscore_vision/benchmarks/laion_fmri/__init__.py
+++ b/brainscore_vision/benchmarks/laion_fmri/__init__.py
@@ -33,7 +33,7 @@ explicitly here.
 """
 
 from brainscore_vision import benchmark_registry
-from .benchmark import LAIONfMRI, LAIONfMRIRSA
+from .benchmark import LAIONfMRI, LAIONfMRICKA, LAIONfMRIRSA
 
 # ── Per-subject pool ridgecv variants (HEADLINE) ──────────────────────────
 # Each subject's 5,833-image pool (1,121 shared + 4,712 subject-unique).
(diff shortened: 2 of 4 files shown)


KartikP:
retest this please

PR #2486, opened 2026-08-12
feat(models): add regnety_032 (timm ra_in1k)

KartikP:
## Model

`regnety_032` from timm, tag `regnety_032.ra_in1k` (RandAugment recipe, ImageNet top-1 ~82.0 — the strongest of the three available tags for this architecture). 19.4M params, 224x224.

RegNetY is near-absent from the leaderboard: only `regnet_y_400mf` and a `regnet_y_128gf_E2E` are registered today, so a mid-size member fills a genuine gap in architecture coverage rather than adding another point near an existing cluster.

## Layers

```
['stem', 's1', 's2', 's3', 's4', 'final_conv', 'head']
```

All seven verified present on the instantiated graph.

`behavioral_readout_layer='head'` is set **explicitly**. `ModelCommitment` otherwise falls back to `layers[-1]`, which would make the readout silently depend on list order — and the readout drives the behavioral benchmarks.

## Preprocessing

`ra_in1k` expects standard ImageNet normalisation, which is exactly what `load_preprocess_images` applies by default `(0.485, 0.456, 0.406) / (0.229, 0.224, 0.225)`, so no override is needed.

## Checks run

- registers in `model_registry`
- `ImportPlugin.locate_plugin()` resolves it (the registration is a single literal `model_registry['regnety_032']` line, which is what discovery greps for)
- every declared layer exists on the module graph
- weights not downloaded locally; the plugin test exercises the real load in CI


KartikP:

KartikP:

PR #2511, opened 2026-09-05
Include preprocessing fingerprint in activation cache key

### LINKED ISSUE #2503: Activation cache key omits preprocessing image_size: identical identifiers with different input resolutions silently reuse stale activations
**Version / environment**

- `brainscore_vision` 2.3.22, repo at commit `cc394fa`; `result_caching` 0.5
- Python 3.11.9, torch 2.13.0+cpu, Windows 11, CPU only
- Benchmark: `FreemanZiemba2013public.V1-pls`
- `BRAINSCORE_CACHE_PLUGIN_REVISION` unset (the default)

**Summary**

The memoization key for stored activations does not include the preprocessing input
resolution. If two `ModelCommitment`s share an `identifier` but differ in `image_size`, the
second silently reuses the first one's cached activations. No warning is emitted, the
preprocessing function is never called, and the benchmark returns a plausible score for a
resolution the model never saw.

This matters for anyone evaluating several input-resolution variants of the same model
locally, which is a natural way to probe how sensitive a score is to preprocessing. Everything
below is local scoring; I have not examined the submission pipeline and make no claim about
it either way.

**Relation to the recent cache-key work**

This sits beside #2472 and its follow-ups (#2473, #2476, #2478, #2480, #2487, #2488, #2489).
`model_helpers/activations/cache_key.py` names this hazard directly — a plugin "can be revised
to load different weights *or apply different preprocessing* while still registering as
`alexnet`". What that machinery covers is a *committed* plugin change, when opted in, which
leaves the case here open on three counts:

- it is off by default (`BRAINSCORE_CACHE_PLUGIN_REVISION`), so local runs key on bare
  identifiers — #2472 puts it as "Unset, keys are byte-for-byte what they were before this
  module existed and the plugin lookup is not even attempted";
- `git log -1 -- <dir>` is blind to uncommitted working-tree edits, which is how resolution
  variants are usually tried. #2472 gives this as a reason for making revisioning opt-in: it
  "churns on commit while staying blind to the edit actually under test";
- it distinguishes revisions *over time*, not two configurations coexisting now. No test in
  `tests/test_model_helpers/activations/test_cache_key.py` exercises a preprocessing or
  `image_size` difference.

So this is a gap beside that work rather than a regression in it. (For completeness: the repro
below would not be cached at all under `BRAINSCORE_CACHE_PLUGIN_REVISION=1`. `collision-test`
is not a registered plugin, so `model_cache_identifier('collision-test')` returns `None` and
#2487's refusal path applies, where a registered `alexnet` resolves to `alexnet@ce04678279e9`.
The collision needs the default configuration.)

**Reproduction**

`repro_cache_image_size.py`, below — self-contained, points `RESULTCACHING_HOME` at a fresh
temp dir, so it neither reads nor writes an existing `~/.result_caching`. Numbers below are
from a run starting at zero cache entries. Two `ModelCommitment`s around torchvision AlexNet,
`region_layer_map={'V1': 'features.2'}`, `visual_degrees=8`, differing only in the `image_size`
bound into `functools.partial(load_preprocess_images, ...)` and in the identifier:

| run | identifier | image_size | preprocessing called | ceiled score | new activation cache entries |
|---|---|---|---|---|---|
| 1a | `collision-test` | 224 | yes, `(64, 3, 224, 224)` | 0.2617980818943393 | 1 |
| 1b | `collision-test` | 112 | **no** | 0.2617980818943393 | 0 |
| 2a | `collision-test-res224` | 224 | yes, `(64, 3, 224, 224)` | 0.2617980818943393 | 1 |
| 2b | `collision-test-res112` | 112 | yes, `(64, 3, 112, 112)` | 0.24269800129223137 | 1 |

Run 1b returns 1a's score exactly, to all digits, and its preprocessing function is never
entered. Runs 2a and 2b, which differ only in identifier, produce different scores as
expected. (1a additionally writes the `_place_on_screen`, ceiling and assembly entries, being
the first run against an empty cache.)

Wall-clock time does not give it away either. Taking 2a as the reference for a genuine 224 run,
the forward pass is a small share of it — 4 s of activations against 58 s of cross-validation,
63.4 s total — and cross-validation is not memoized, so it reruns in full every time. What
makes this worse than merely unnoticeable is the comparison across all three:

| run | activations | cross-validation | total |
|---|---|---|---|
| 2a — genuine 224 | 4 s | 58 s | 63.4 s |
| 1b — silently reused 224 | none | 60 s | 60.6 s |
| 2b — genuine 112 | 1 s | 25 s | 27.6 s |

A correct 112 run is less than half the duration because the feature map, and with it the
regression, shrinks: the assemblies these runs cached hold 46,656 neuroids for `features.2` at
224 and 10,816 at 112. So 1b does not merely fail to look suspicious: it
lands within 5% of the 224 run it actually was, and a resolution change that *did* take effect
would have been the conspicuous one.

<details>
<summary><code>repro_cache_image_size.py</code></summary>

```python
"""Minimal repro: the stored-activation cache key ignores preprocessing image_size.

Two ModelCommitments that share an identifier but differ in image_size: the
second serves the first one's cached activations, and its preprocessing is
never called.

Run:
    python repro_cache_image_size.py

Uses an isolated RESULTCACHING_HOME (a fresh temp dir), so it cannot read or
write an existing ~/.result_caching.
"""
import functools
import os
import tempfile
import time

CACHE = os.environ.setdefault(
    'RESULTCACHING_HOME', os.path.join(tempfile.gettempdir(), 'bs_repro_cache'))
# must be set before brainscore/result_caching import: the storage directory is
# read when the @store_xarray decorator is instantiated at import time.

import torchvision  # noqa: E402
from brainscore_vision import load_benchmark  # noqa: E402
from brainscore_vision.model_helpers.activations.pytorch import (  # noqa: E402
    PytorchWrapper, load_preprocess_images)
from brainscore_vision.model_helpers.brain_transformation import ModelCommitment  # noqa: E402

BENCHMARK = 'FreemanZiemba2013public.V1-pls'


def cache_files():
    return {os.path.join(root, f)
            for root, _, files in os.walk(CACHE) for f in files}


def run(label, identifier, image_size, benchmark):
    seen = []
    base = functools.partial(load_preprocess_images, image_size=image_size)

    def preprocessing(image_filepaths):
        out = base(image_filepaths)
        seen.append(tuple(out.shape))
        return out

    wrapper = PytorchWrapper(identifier=identifier,
                             model=torchvision.models.alexnet(pretrained=True),
                             preprocessing=preprocessing)
    commitment = ModelCommitment(identifier=identifier, activations_model=wrapper,
                                 layers=['features.2'],
                                 region_layer_map={'V1': 'features.2'},
                                 visual_degrees=8)
    before = cache_files()
    t0 = time.time()
    score = benchmark(commitment)
    elapsed = time.time() - t0
    new = sorted(os.path.basename(p) for p in cache_files() - before)

    print(f"\n--- {label}: identifier={identifier!r} image_size={image_size} ---")
    print(f"  preprocessing called : {'yes, ' + str(seen[0]) if seen else 'NO -- served from cache'}")
    print(f"  ceiled score         : {float(score.values)!r}")
    print(f"  elapsed              : {elapsed:.1f}s")
    print(f"  new cache entries    : {len(new)}")
    for name in new:
        print(f"      {name}")
    return float(score.values)


def main():
    print(f"RESULTCACHING_HOME = {CACHE}")
    print(f"cache entries at start: {len(cache_files())}")
    benchmark = load_benchmark(BENCHMARK)

    a = run('1a', 'collision-test', 224, benchmark)
    b = run('1b', 'collision-test', 112, benchmark)
    c = run('2a', 'collision-test-res224', 224, benchmark)
    d = run('2b', 'collision-test-res112', 112, benchmark)

    print("\n=== result ===")
    print(f"same identifier,      224 vs 112 : {a!r} vs {b!r} -> "
          f"{'IDENTICAL (collision)' if a == b else 'differ'}")
    print(f"distinct identifiers, 224 vs 112 : {c!r} vs {d!r} -> "
          f"{'identical' if c == d else 'differ (expected)'}")


if __name__ == '__main__':
    main()
```

</details>

**Why it happens**

The key is readable directly off disk — it is the cache filename. From run 1a:

```
identifier=collision-test,stimuli_identifier=FreemanZiemba2013.aperture-public--target8.00--source4.00,number_of_trials=1,require_variance=False.pkl
```

`visual_degrees` is present, inside the stimulus identifier; `image_size` is absent. Run 1b
requests 112 and produces this same string, so it hits. Why the key is shaped that way:

- `ActivationsExtractorHelper._from_paths_stored`
  (`brainscore_vision/model_helpers/activations/core.py:124-127`) declares `identifier`,
  `stimuli_identifier`, `number_of_trials`, `require_variance` as parameters; `stimuli_paths`
  and `layers` are explicitly ignored for the key.
- `result_caching` builds the key from the declared call arguments — `_Storage.__call__` calls
  `getcallargs` and `get_function_identifier` (`result_caching/__init__.py`, `__call__` /
  `getcallargs` / `get_function_identifier`; separate repo, so line numbers are not pinned by
  `cc394fa`).
- `image_size` is not a parameter of that call. It is bound into the preprocessing closure
  (`functools.partial(load_preprocess_images, image_size=...)`, e.g.
  `brainscore_vision/models/alexnet/model.py:25`) and so cannot appear in the key by
  construction.
- By contrast, `visual_degrees` *does* reach the key, indirectly, through the
  `--target8.00--source4.00` suffix that `_place_on_screen` appends to the stimulus set
  identifier (`brainscore_vision/benchmark_helpers/screen.py:52`).

So one of the two parameters that determine the effective input resolution is in the key and
the other is not.

**Suggested fixes**, in increasing order of invasiveness:

1. Document it — a note in `docs/source/modules/developer_clarifications.rst` item 2, next to
   the existing result_caching material, stating that an identifier must be unique per
   preprocessing configuration.
2. Assert — record a hash of the preprocessing configuration alongside the cached entry and
   raise when it differs from the current one for the same identifier.
3. Key on it — make the resolution (or a preprocessing fingerprint) an explicit parameter of
   the memoized call. The value is mostly already at hand: 490 of the 562 model plugin
   directories set a `.image_size` attribute on their wrapper. But nothing outside
   `brainscore_vision/models/` reads that attribute today, so nothing validates it against the
   preprocessing actually in effect — a plugin whose attribute has drifted would yield a wrong
   but stable-looking key. The values would need checking against the effective preprocessing
   before they could serve as key material.

   (Both counts are per plugin *directory*; the registry exposes more identifiers than there
   are directories, since a directory can register several.)

Option 1 alone would remove the silent part, which is the actual problem here.

**Workaround**

Use a distinct identifier per resolution variant. The key is built from the extractor's
`self.identifier` (`core.py:96`), so the identifier has to be set on the `PytorchWrapper`
itself, not only on the `ModelCommitment`.

I'm happy to open a PR for option 1 — the note in
`docs/source/modules/developer_clarifications.rst`, item 2, alongside the existing
result_caching material — if that's the direction you'd prefer.

nilsleut:
The stored-activation key was built from identifier, stimuli_identifier, number_of_trials and require_variance only. Preprocessing resolution is bound into the preprocessing closure and could not reach the key, so two extractors sharing an identifier but differing in input resolution served each other's cached activations silently.

Probe the preprocessing callable once per extractor with a synthetic image and pass the resulting shape/dtype fingerprint as an explicit parameter of _from_paths_stored. Extractors whose preprocessing cannot be probed key on 'unprobed'.

Fixes #2503

### DIFF
--- a/brainscore_vision/model_helpers/activations/core.py
+++ b/brainscore_vision/model_helpers/activations/core.py
@@ -1,6 +1,7 @@
 import copy
 import os
 import cv2
+import re
 import tempfile
 from typing import Dict, Tuple, List, Union
 
@@ -10,6 +11,7 @@ from collections import OrderedDict
 from multiprocessing.pool import ThreadPool
 
 import numpy as np
+from PIL import Image
 from tqdm.auto import tqdm
 import xarray as xr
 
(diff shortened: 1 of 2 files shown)


KartikP (approved):
This is really strong work! Thank you so much.

One note, that I'll also make is that `temporal/core/extractor.py:120` has its own `_from_paths_stored` with the same `@store_xarray` and no resolution component, so a future video model (that doesn't properly exist) still will have the gap.

That said, this can be a follow-up.

